Issue #836, opened 2026-05-31
fix(automation/ci_driver): 'pushed CI fixes' logged even when no commit was made and remote is unchanged


mvillmow:
## Problem

After the pre-sync (#832) the worktree's HEAD matches ``origin/<pr-head>``. If the Claude/Codex session returns without making a new commit (because nothing changed, because the resumed session believed the fix was already applied, because the session timed out before committing, etc.) HEAD remains equal to ``origin/<pr-head>``. The subsequent ``git push --force-with-lease=<branch> origin HEAD:<branch>`` is a **silent no-op that exits 0**.

``_run_ci_fix_session`` interprets the 0 exit as success and logs:

```
Issue #N: pushed CI fixes for PR #M
Issue #N: CI fix applied successfully (attempt 1)
Issue #N: CI drive completed
```

But the remote PR is unchanged — the same commit it had before the run. The driver reports success while doing nothing.

## Reproduction

Run ``20260531T085744Z`` against ProjectOdyssey:

- Driver log lines:
  ```
  Issue #5450: pushed CI fixes for PR #5485
  Issue #5450: CI fix applied successfully (attempt 1)
  ```
- Actual state of PR #5485 a few hours later: head SHA ``ab56819d``, committed ``2026-05-30T19:46:53Z`` — *yesterday*, before the run started.

Repeats for #5135 / PR #5473 (tip ``6f407dc1`` committed ``2026-05-30T19:31:39Z``), #5451 / PR #5487 (tip ``60e6c1a0`` committed ``2026-05-31T04:08:55Z`` — pre-run), and so on. The post-sync state and a session that returns without committing exactly hide this from the success path.

## Fix

Detect "no commits to push" between the sync and the push, and treat it as fix-failed-this-iteration:

```python
# Before invoking the agent, snapshot the synced HEAD.
pre_agent_sha = run(["git", "rev-parse", "HEAD"], cwd=worktree_path).stdout.strip()

# ... agent session ...

post_agent_sha = run(["git", "rev-parse", "HEAD"], cwd=worktree_path).stdout.strip()
if post_agent_sha == pre_agent_sha:
    logger.warning(
        "Issue #%s: agent session produced no new commit on top of "
        "origin/%s (HEAD unchanged at %s); nothing to push",
        issue_number,
        pr_head_branch,
        pre_agent_sha[:8],
    )
    return False
```

[...]

Issue #832, opened 2026-05-31
fix(automation/ci_driver): CI fix push targets the wrong branch (Claude switches branches, push follows HEAD)


mvillmow:
## Problem

`ci_driver` checks out the PR's head branch in a worktree, then hands Claude a prompt that includes the working directory but does NOT say "stay on this branch". Claude, following the project's general workflow guidance in `CLAUDE.md` (the "Git Workflow" section that says `git checkout -b <issue-number>-description`), often creates a brand-new feature branch and commits there instead.

Then the driver runs:

```
git push origin HEAD
```

Because `HEAD` now points to Claude's new branch (e.g. `5391-fix-ci`), this pushes to a **new remote branch with that name** — NOT to the PR's head branch (`5391-auto-impl`). The push succeeds (the new remote ref doesn't exist, so it's fast-forward), and the driver logs:

```
Issue #5391: pushed CI fixes for PR #5483
Issue #5391: CI fix applied successfully (attempt 1)
```

But PR #5483 is unchanged. Its head branch is still at the old commit; only a stray `5391-fix-ci` branch exists on the remote that no PR points at.

## Reproduction

Run `~/drive-prs-green-ecosystem.sh`. In run `20260531T060822Z` against ProjectOdyssey the driver claimed to push fixes for 11 PRs. Spot check #5483 / issue #5391:

```
# What the driver said:
[INFO] _review_utils: Found PR #5483 for issue #5391 via branch name
[INFO] worktree_manager: Branch 5391-auto-impl already exists, reusing it
[INFO] ci_driver: Issue #5391: pushed CI fixes for PR #5483
[INFO] ci_driver: Issue #5391: CI fix applied successfully (attempt 1)
```

```
# What actually happened on the remote:
$ gh pr view 5483 --json headRefName,headRefOid
head: 5391-auto-impl  oid: 89f2b646  (unchanged — last commit 2026-05-29)

$ gh api '/repos/HomericIntelligence/ProjectOdyssey/branches?per_page=100' \
    | jq '.[] | select(.name | test("5391"))'
5391-auto-impl → 89f2b646   ← PR head, unchanged
5391-fix-ci → f0b680e2     ← STRAY branch from today's run, no PR points here
```

Local reflog confirms:

```
$ git reflog 5391-fix-ci
f0b680e2 5391-fix-ci@{0}: commit: fix: Address CI failures for PR ProjectOdyssey#5483
[...]

PR #843, opened 2026-05-31
fix(automation/session_naming): drop githash from automation-loop sessions

### LINKED ISSUE #841: fix(automation/session_naming): drop githash from ALL automation-loop sessions; keep issue-or-PR scope only
## Problem

``session_name(repo, issue, agent, githash)`` and its UUIDv5 derivation include the current ``main`` githash. Every agent in the automation pipeline — planner, plan_reviewer, implementer, address_review, pr_reviewer, ci_driver — feeds the live ``current_trunk_githash`` into the session-naming tuple. Whenever ``main`` advances between runs, the deterministic session UUID for a given (repo, issue, agent) tuple changes, and ``invoke_claude_with_session`` falls back to a fresh ``--session-id`` create rather than ``--resume``\-ing the prior transcript.

For the entire automation loop this is the wrong scope:

- The issue (and the PR that closes it) is the long-lived artifact. Its state evolves across many main-bumps.
- The loop's job is to incrementally make progress on that artifact. The most useful context for the next iteration is what the previous iteration tried, learned, and committed — not "we were at githash X when we started".
- The current scoping fragments every agent's transcripts across N sessions per main-bump, none of which can learn from each other.

The user's mandate: **the session id must be PR-or-issue scoped, the same for every stage of the automation loop, never commit/hash based**.

## Fix

Drop ``githash`` from the session-naming tuple entirely. The same (repo, issue, agent) tuple resolves to the same session UUID across all main-bumps, for every agent in the loop.

### Files to change

- ``hephaestus/automation/session_naming.py``
  - ``session_name(repo, issue, agent)`` (3-tuple, githash removed)
  - ``session_uuid_for(repo, issue, agent)`` (delegates to the new ``session_name``)
  - ``current_trunk_githash`` itself can stay if anything else depends on it for non-session purposes; if not, delete it.

- All callers of ``invoke_claude_with_session`` in:
  - ``planner_claude.py``
  - ``plan_reviewer.py``
  - ``implementer_phase_runner.py`` (3 call sites)
  - ``address_review.py``
  - ``pr_reviewer.py``
  - ``ci_driver.py`` (3 call sites: advise, fix session, drive-green learnings)

  Each currently has the pattern::

      githash = current_trunk_githash(self.repo_root)
      invoke_claude_with_session(..., githash=githash, ...)

  Becomes::

      invoke_claude_with_session(..., ...)

  with ``githash`` removed from ``invoke_claude_with_session``'s signature.

### Risk and mitigations

Same as previously listed (Claude can carry stale assumptions across rebases) — and the three already-landed fixes (#823, #832, #836) bound the risk:

- The per-iteration prompt includes the *current* failing CI logs and issue body, so the model sees the live state regardless of transcript history.
- The pre-agent worktree sync (#832) resets HEAD to the actual PR head before every iteration.
- The "HEAD didn't advance" guard (#836) catches no-op sessions, so a confused-resume produces an honest failure instead of a silent no-op push.

Applied uniformly across the loop, these turn the staleness risk into "the session has more context than the current state, occasionally" — which is the desirable property, since the alternative is "the session has zero context every time main moves".

## Acceptance

- ``session_name`` and ``session_uuid_for`` no longer take a ``githash`` argument.
- ``invoke_claude_with_session`` no longer takes a ``githash`` argument.
- Every caller in the automation loop is updated; ``current_trunk_githash`` is no longer called for session-naming anywhere.
- For a given (repo, issue, agent) tuple, the session UUID is invariant across main-bumps.
- New unit tests verify: the UUID depends only on (repo, issue, agent); calling with the same tuple twice returns the same UUID; different agents on the same issue produce different UUIDs; same agent on different issues produces different UUIDs.
- A re-run of any automation-loop stage on the same issue resumes the prior transcript rather than creating a fresh session.

## Related

- #823 — dot-encoding fix on ``session_jsonl_path``. Made resume reliable; this makes resume *useful* (transcripts now actually persist across main-bumps).
- #832, #836 — bound the staleness risk; see "Mitigations".
- #840 — ``/learn`` on detected merge. Both this and #840 push the loop toward PR-scoped, merge-aware bookkeeping.

mvillmow:
## Summary

Every agent in the pipeline (planner, plan_reviewer, implementer, address_review, pr_reviewer, ci_driver) fed the live `current_trunk_githash` into the session-naming tuple. The deterministic UUID changed whenever `main` advanced, so each main-bump forked a fresh session family and ``invoke_claude_with_session`` fell back to a ``--session-id`` create rather than ``--resume``-ing the prior session.

For the whole automation loop this is the wrong scope: the long-lived artifact is the issue/PR, not the commit at which the loop happened to start. The same (repo, issue, agent) tuple should resolve to the same session id across every iteration, every run, and every main-bump for as long as the artifact exists.

## Changes

- ``session_name`` and ``session_uuid`` in ``session_naming.py``: ``githash`` parameter removed.
- ``invoke_claude_with_session`` in ``claude_invoke.py``: ``githash`` parameter removed.
- All 8 call sites updated: ``planner_claude``, ``plan_reviewer``, ``pr_reviewer``, ``address_review``, ``implementer_phase_runner`` (3 sites), ``ci_driver`` (3 sites).
- Unused ``current_trunk_githash`` imports removed from those modules. The helper itself is kept for callers that still need a trunk-SHA observation for non-session purposes (logging, state files).
- A new regression test (`test_signature_does_not_accept_githash_kw`) makes ``githash=`` a hard ``TypeError`` so the invariant cannot silently regress via a future caller.

## Risk and mitigations

Three already-landed fixes bound the staleness risk:

- **#823 (dot-encoding fix)**: made ``--resume`` reliable in the first place.
- **#832 (worktree pre-sync)**: every iteration starts from the PR's actual remote head so the workspace state always matches live.
- **#836 (no-commit guard)**: confused-resume produces an honest failure instead of a silent no-op push.

Together these turn the staleness concern into "the session has more context than the current state, occasionally" — which is the desired property, since the alternative was "the session has zero context every time main moves".

## Test plan

- [x] `pytest tests/unit/automation` (938 passed)
- [x] Updated `test_session_naming.py` to drop the githash positional, kept all distinctness assertions, added the kwarg-rejection regression test.
- [x] Updated `test_claude_invoke_session.py`: replaced `test_different_githash_starts_fresh_family` with `test_session_id_is_githash_invariant` (the inverse assertion).
- [x] Updated `test_pr_reviewer_posting.py` and `test_address_review.py` to drop now-irrelevant `current_trunk_githash` patches.
- [x] Updated `test_ci_driver.py` to drop two `current_trunk_githash` patches that referenced the removed import.
- [x] `pre-commit run` clean (ruff format auto-applied; re-tested green).

Closes #841

### DIFF
--- a/hephaestus/automation/address_review.py
+++ b/hephaestus/automation/address_review.py
@@ -58,7 +58,7 @@ from .github_api import (
 )
 from .models import AddressReviewOptions, ReviewPhase, ReviewState, WorkerResult
 from .prompts import get_address_review_prompt
-from .session_naming import AGENT_IMPLEMENTER, current_trunk_githash
+from .session_naming import AGENT_IMPLEMENTER
 from .status_tracker import StatusTracker  # noqa: F401 — re-exported for test patching
 from .worktree_manager import WorktreeManager  # noqa: F401 — re-exported for test patching
 
@@ -176,13 +176,11 @@ def run_address_fix_session(
             )
             return parsed
 
-        githash = current_trunk_githash(repo_root)
         repo_slug = get_repo_slug(repo_root)
         stdout, _ = invoke_claude_with_session(
             repo=repo_slug,
             issue=issue_number,
             agent=AGENT_IMPLEMENTER,
-            githash=githash,
             prompt=prompt,
             model=implementer_model(),
             cwd=worktree_path,
--- a/hephaestus/automation/ci_driver.py
+++ b/hephaestus/automation/ci_driver.py
@@ -48,7 +48,7 @@ from .git_utils import (
 from .github_api import _gh_call, gh_issue_json, gh_pr_checks
 from .models import CIDriverOptions, WorkerResult
 from .prompts import get_advise_prompt
-from .session_naming import AGENT_ADVISE, AGENT_CI_DRIVER, current_trunk_githash
+from .session_naming import AGENT_ADVISE, AGENT_CI_DRIVER
 from .status_tracker import StatusTracker
 from .worktree_manager import WorktreeManager
 
@@ -514,13 +514,11 @@ class CIDriver:
                     sandbox="read-only",
                 )
                 return (result.stdout or "").strip()
-            githash = current_trunk_githash(self.repo_root)
             repo_slug = get_repo_slug(self.repo_root)
             stdout, _ = invoke_claude_with_session(
                 repo=repo_slug,
                 issue=issue_number,
                 agent=AGENT_ADVISE,
-                githash=githash,
                 prompt=prompt,
                 model=advise_model(),
                 cwd=self.repo_root,
@@ -958,14 +956,12 @@ class CIDriver:
             # independent of the implementer's transcript. The first fix call
             # creates it via --session-id; later calls resume it. The codex
             # path above instead resumes the raw ``session_id`` it was handed.
-            githash = current_trunk_githash(self.repo_root)
             repo_slug = get_repo_slug(self.repo_root)
             try:
                 stdout, _ = invoke_claude_with_session(
                     repo=repo_slug,
                     issue=issue_number,
                     agent=AGENT_CI_DRIVER,
-                    githash=githash,
                     prompt=prompt,
                     model=implementer_model(),
                     cwd=worktree_path,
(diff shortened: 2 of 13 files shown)


mvillmow:
Closing in favor of a freshly-rebased PR — #843 hit a merge conflict in tests/unit/automation/test_ci_driver.py after #844 merged. The reopened PR contains identical content rebased onto current main.

PR #943, opened 2026-06-05
docs(readme): point non-Linux users to wheel install path

### LINKED ISSUE #767: [MINOR] Install/upgrade docs for non-Linux platforms missing (pixi linux-64 only, wheel pure-Python)
## Finding

Workspace platforms=['linux-64'] is the only supported pixi platform, yet the published wheel is pure-Python and the project classifies for 3.10-3.13. Install/upgrade docs for non-Linux platforms (Windows/macOS) aren't explicit — users on macOS/Windows can pip install the wheel but cannot pixi install the dev env.

**Location:** `pixi.toml:5, pyproject.toml`

## Proposed fix

Document the asymmetry in README or CONTRIBUTING.md: 'macOS/Windows: use pip install; dev environment is Linux-only via pixi.'

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S12

---
Tracking: #708

mvillmow:
## Summary

Add explicit platform support guidance to README: pixi dev environment is Linux-64 only, while the published wheel supports macOS/Windows via pip. Cross-reference the canonical platform comparison table in CONTRIBUTING.md (from PR #932). Add regression test to prevent anchor drift between README and CONTRIBUTING.

Closes #767

## Test plan

- [x] Run regression tests: `pixi run pytest tests/unit/validation/test_readme_platform_support.py -v` (all 3 pass)
- [x] Run full validation suite: `pixi run pytest tests/unit/validation -v` (649 passed)
- [x] Pre-commit hooks pass: `pre-commit run --files README.md tests/unit/validation/test_readme_platform_support.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,48 @@ codex plugin add hephaestus@project-hephaestus
 
 The Codex manifest lives in [`.codex-plugin/plugin.json`](.codex-plugin/plugin.json), and the marketplace entry lives in [`.agents/plugins/marketplace.json`](.agents/plugins/marketplace.json).
 
+### Installing in Another Project
+
+ProjectHephaestus is published to PyPI as `homericintelligence-hephaestus`.
+The wheel is pure-Python and installs on Linux, macOS, and Windows
+(see `requires-python` in [`pyproject.toml`](pyproject.toml)). This is
+the supported install path for non-Linux platforms.
+
+**Using pip:**
+
+```bash
+pip install homericintelligence-hephaestus
+```
+
+**Using Pixi:**
+
+Add to `pyproject.toml`:
+
+```toml
+[project]
+dependencies = [
+    "homericintelligence-hephaestus>=0.9,<1",
+]
+```
+
+Or add a PyPI entry to `pixi.toml`:
+
+```toml
+[pypi-dependencies]
+homericintelligence-hephaestus = ">=0.9,<1"
+```
+
+Then run `pixi install` to resolve the dependency.
+
+After 1.0 ships, bump these constraints to `>=1.0,<2`.
+
+**For local development (path dependency):**
+
+```toml
+[pypi-dependencies]
+homericintelligence-hephaestus = { path = "../ProjectHephaestus", editable = true }
+```
+
 ## Key Features
 
 ### General Utilities (`hephaestus.utils`)
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ ProjectHephaestus/
 
 This project uses [Pixi](https://pixi.sh) for environment management, which automatically handles dependencies and creates isolated environments.
 
+> **Platform note:** The pixi developer environment is **Linux-64 only** (see
+> `platforms` in [`pixi.toml`](pixi.toml)). On macOS or Windows, install the
+> published wheel into a plain virtualenv instead — see
+> [From PyPI](#from-pypi) above. The
+> full comparison table (install paths, supported platforms, Python versions)
+> lives in [CONTRIBUTING.md#platform-support](CONTRIBUTING.md#platform-support).
+
 ### Prerequisites
 
 Install Pixi by following the [official installation guide](https://pixi.sh/install/).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -136,6 +143,9 @@ The Codex manifest lives in [`.codex-plugin/plugin.json`](.codex-plugin/plugin.j
 ### Installing in Another Project
 
 ProjectHephaestus is published to PyPI as `homericintelligence-hephaestus`.
+The wheel is pure-Python and installs on Linux, macOS, and Windows (Python
```
mvillmow: Hardcodes `3.10+` inline. The plan explicitly warned against hardcoded version numbers because they drift when `requires-python` is bumped. Drop the version claim and keep only the pointer: `(see \`requires-python\` in [\`pyproject.toml\`](pyproject.toml))`.

tests/unit/validation/test_readme_platform_support.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_readme_flags_pixi_as_linux_only() -> None:
+    """README must warn that the pixi dev env is Linux-only before `pixi install`."""
+    text = README.read_text(encoding="utf-8")
+    pixi_section_start = text.index("## Getting Started with Pixi")
+    pixi_install_pos = text.index("pixi install", pixi_section_start)
+    preface = text[pixi_section_start:pixi_install_pos]
+    assert "linux-64" in preface.lower() or "linux only" in preface.lower(), (
+        "README must mention the linux-64 restriction before the first "
```
mvillmow, line 50: `text.index("pixi install", pixi_section_start)` raises ValueError if the literal string moves or is renamed, producing a confusing test error rather than a clear assertion failure. Wrap in a try/except or use `find()` and assert the result is not -1 with an explicit message.

mvillmow (commented):
POLICY VIOLATION: Closes, auto-merge, signed-commits — PR body is empty (no Closes #767), auto-merge is prematurely enabled, and no signed commits are present in the signing state array.

mvillmow (commented):
POLICY VIOLATION: Closes, auto-merge, signed-commits — PR description is empty (no Closes #767), auto_merge_enabled=true during review, and commit signing state array is empty (no verified signatures).

README.md
mvillmow: Fixed: the "Installing in Another Project" section has been restored to README.md without the hardcoded `3.10+` version claim. The line now reads: `(see \`requires-python\` in [\`pyproject.toml\`](pyproject.toml))` — a pure semantic reference that stays correct when `requires-python` is bumped. Commit 774b656.

tests/unit/validation/test_readme_platform_support.py
mvillmow: Fixed: both `text.index()` calls in `test_readme_flags_pixi_as_linux_only` are now replaced with `text.find()` + explicit `assert != -1`. The `pixi install` lookup was converted in 0a9b0cd; this commit (774b656) applies the same `.find()` + assert pattern to the sibling `## Getting Started with Pixi` heading lookup (line 40), so both produce a clear, readable assertion failure message instead of a confusing ValueError if the heading is ever renamed or removed.

mvillmow (commented):
Fixed 2 unresolved threads in commit 774b656:

1. **PRRT_kwDOQww0as6HSvtG** (README.md hardcoded `3.10+`): Restored the "Installing in Another Project" section that was accidentally dropped in the force-push rewrite. The version claim is replaced with a pure semantic reference: `(see `requires-python` in [`pyproject.toml`](pyproject.toml))` — no hardcoded version that can drift.

2. **PRRT_kwDOQww0as6HSvtI** (test `text.index()` raises ValueError): All `text.index()` calls in `test_readme_flags_pixi_as_linux_only` are now replaced with `text.find()` + `assert != -1` with an explicit message. The `pixi install` lookup was converted in 0a9b0cd; this commit applies the same pattern to the `## Getting Started with Pixi` heading lookup, so both failure paths produce clear, readable assertion errors.

All 643 unit tests pass; pre-commit hooks (ruff, mypy, markdownlint) pass on the changed files.

PR #951, opened 2026-06-05
fix(validation): confine markdown link resolution to repo_root

### LINKED ISSUE #752: [MINOR] validation/markdown.py resolves user-supplied link paths without confining to repo_root
## Finding

(source_file.parent / file_part).resolve() doesn't restrict to repo root — a malicious markdown file could check existence of arbitrary host paths via the validator. Read-only impact, but a subtle information-disclosure surface.

**Location:** `hephaestus/validation/markdown.py:164,554`

## Proposed fix

After resolve(), check the result is under repo_root before any .exists() / .is_file() call.

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S08

---
Tracking: #708

mvillmow:
## Summary

Fixes a path-traversal information-disclosure vulnerability in the markdown link validator. The functions `validate_relative_link()` and `validate_internal_link()` now check that resolved link paths remain under `repo_root` before calling `.exists()`, preventing a malicious markdown file from probing arbitrary host filesystem paths using `../` sequences.

## Changes

- **Added `_is_within_repo()` helper**: Private function that validates path containment using `Path.is_relative_to()` after resolving both the candidate and repo_root paths. Handles symlinks and `..` segments correctly.
- **Updated `validate_relative_link()`**: Now calls the boundary check before `.exists()` and returns a distinct error message ("Link target outside repository") for traversal attempts.
- **Updated `validate_internal_link()`**: Applies the same boundary check to both relative and absolute-from-root link branches, with explicit `.resolve()` on the absolute path to ensure proper containment checking.
- **Added 4 new test cases**: Covers traversal rejection (`../../etc/passwd`), absolute traversal (`/../../etc/passwd`), and regression case (legitimate relative links within the repo still validate correctly).

## Testing

All 44 validation tests pass, including the 4 new traversal-rejection tests. Ruff formatting and mypy type-checking pass.

Closes #752

### DIFF
--- a/hephaestus/validation/markdown.py
+++ b/hephaestus/validation/markdown.py
@@ -143,6 +143,8 @@ def _is_within_repo(candidate: Path, repo_root: Path) -> bool:
         True if candidate is the same as or nested under repo_root.
 
     """
+    # resolve() is non-strict (default): confine the target before any
+    # existence check, even if the link target does not exist yet.
     try:
         candidate.resolve().relative_to(repo_root.resolve())
     except ValueError:
--- a/hephaestus/validation/markdown.py
+++ b/hephaestus/validation/markdown.py
@@ -128,6 +128,28 @@ def extract_markdown_links(content: str) -> list[tuple[str, int]]:
     return links
 
 
+def _is_within_repo(candidate: Path, repo_root: Path) -> bool:
+    """Return True iff candidate is within repo_root after resolving both paths.
+
+    Both paths are compared after .resolve() so symlinks and .. segments
+    cannot escape the boundary. Used to prevent user-supplied markdown link
+    targets from probing arbitrary host paths.
+
+    Args:
+        candidate: Path to check for containment.
+        repo_root: Repository root boundary.
+
+    Returns:
+        True if candidate is the same as or nested under repo_root.
+
+    """
+    try:
+        candidate.resolve().relative_to(repo_root.resolve())
+    except ValueError:
+        return False
+    return True
+
+
 def validate_relative_link(
     link: str, source_file: Path, repo_root: Path
 ) -> tuple[bool, str | None]:
@@ -160,8 +182,10 @@ def validate_relative_link(
     if not file_part:
         return True, None
 
-    # Resolve relative path
+    # Resolve relative path and confine to repo_root
     link_path = (source_file.parent / file_part).resolve()
+    if not _is_within_repo(link_path, repo_root):
+        return False, f"Link target outside repository: {link}"
 
     # Check if file exists
     if not link_path.exists():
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
POLICY VIOLATION: Closes #N (PR body empty), auto-merge (enabled=true before GO), signed-commits (empty signing state — no verified commits)

mvillmow (commented):
Correct, complete, fail-closed fix confining both link-resolution sites to repo_root before .exists(); 4 adversarial+positive tests. Only minor gaps: no symlink-escape test, strict=False intent implicit.

tests/unit/validation/test_validation.py
```diff
@@ -445,6 +445,49 @@ def test_validate_all_links_empty_dir(self, tmp_path):
...
+        """Legitimate relative links inside repo_root continue to validate."""
+        from hephaestus.validation.markdown import validate_relative_link
+
+        md_file = tmp_path / "docs" / "a" / "index.md"
+        md_file.parent.mkdir(parents=True)
+        md_file.write_text("# Index")
+        target = tmp_path / "docs" / "b.md"
+        target.write_text("# B")
+        valid, error = validate_relative_link("../b.md", md_file, tmp_path)
+        assert valid is True
```
mvillmow, line 488: Consider adding one test that pins symlink-escape behavior: create an in-repo symlink (e.g. `tmp_path/docs/escape` → `/etc`) and assert `validate_internal_link("escape/passwd", md_file, tmp_path)` is rejected with "outside repository". `.resolve()` follows symlinks, so this case is already handled correctly by `_is_within_repo`, but it's currently unverified — the strict plan review flagged it as the one untested vector in this otherwise-thorough suite.

hephaestus/validation/markdown.py
```diff
@@ -128,6 +128,28 @@ def extract_markdown_links(content: str) -> list[tuple[str, int]]:
...
+    Args:
+        candidate: Path to check for containment.
+        repo_root: Repository root boundary.
+
+    Returns:
+        True if candidate is the same as or nested under repo_root.
+
+    """
+    try:
+        candidate.resolve().relative_to(repo_root.resolve())
```
mvillmow, line 149: `candidate.resolve()` relies on the default `strict=False`, which is the intended behavior here (the link target may not exist yet — you want to reject boundary-escape *before* the existence check). A one-line comment stating that non-strict resolution is deliberate would make the intent explicit for future readers; behavior as written is correct.

mvillmow (commented):
Correct, complete, fail-closed fix confining both link-resolution sites to repo_root before .exists(); 4 adversarial+positive tests. Minor: no symlink-escape test in this diff, strict=False intent implicit. Not blocking.

tests/unit/validation/test_validation.py
mvillmow: Addressed: added `test_validate_internal_link_rejects_symlink_escape` (tests/unit/validation/test_validation.py:491) — creates an in-repo `docs/escape` -> `/etc` symlink and asserts `validate_internal_link("escape/passwd", ...)` is rejected with "outside repository". Skips gracefully where symlinks are unsupported.

hephaestus/validation/markdown.py
mvillmow: Addressed: added an inline comment in `_is_within_repo` documenting that `resolve()` is deliberately non-strict (default) so the boundary is confined before any existence check, even when the link target does not yet exist.

PR #953, opened 2026-06-05
fix(nats): remove redundant asyncio.wait_for around next_msg

### LINKED ISSUE #753: [MINOR] nats/subscriber.py double timeout (sub.next_msg(0.5) wrapped in asyncio.wait_for(1.0))
## Finding

Double timeout: 'sub.next_msg(timeout=0.5)' is already polling, then wrapped in 'asyncio.wait_for(..., timeout=1.0)'. The inner timeout always trips first, so the outer wait_for is dead code (or a subtle bug if next_msg's timeout semantics change).

**Location:** `hephaestus/nats/subscriber.py:318-321`

## Proposed fix

Drop the outer asyncio.wait_for; the inner timeout=0.5 is sufficient.

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S09

---
Tracking: #708

mvillmow:
## Summary

Removes the redundant outer `asyncio.wait_for(timeout=1.0)` wrapper around `sub.next_msg(timeout=0.5)` in the NATS polling loop. The inner 0.5s timeout always fires first, making the outer wrapper dead code per issue #753.

Also simplifies the except clause from `(asyncio.TimeoutError, TimeoutError)` to `TimeoutError` since they are equivalent on Python 3.10+ (the project's minimum version).

## Changes

- Removed `asyncio.wait_for()` wrapper from `hephaestus/nats/subscriber.py:318-321`
- Simplified except clause to catch only `TimeoutError`
- Added regression tests in `tests/unit/nats/test_subscriber_polling.py` to prevent re-introduction

## Test Plan

✅ All 61 existing NATS tests pass
✅ 3 new regression tests confirm:
  - `asyncio.wait_for` is not used
  - `next_msg(timeout=0.5)` is called directly
  - Except clause catches only `TimeoutError`
✅ Type checking: Success (305 source files)
✅ Lint: All checks passed

Closes #753

### DIFF
--- a/tests/unit/nats/test_subscriber_polling.py
+++ b/tests/unit/nats/test_subscriber_polling.py
@@ -1,0 +1,37 @@
+"""Tests for the inner polling timeout behavior of _subscribe_loop (#753)."""
+from __future__ import annotations
+
+import inspect
+
+from hephaestus.nats import subscriber
+
+
+def test_polling_loop_source_code_does_not_use_asyncio_wait_for() -> None:
+    """Regression: source code must not contain asyncio.wait_for around next_msg (#753)."""
+    source = inspect.getsource(subscriber.NATSSubscriberThread._subscribe_loop)
+    assert (
+        "asyncio.wait_for" not in source
+    ), "asyncio.wait_for must not be used to wrap next_msg (issue #753)"
+
+
+def test_polling_loop_uses_only_next_msg_timeout() -> None:
+    """The inner timeout=0.5 on next_msg is the only timeout mechanism."""
+    source = inspect.getsource(subscriber.NATSSubscriberThread._subscribe_loop)
+    # Check that next_msg is called with timeout=0.5
+    assert (
+        "next_msg(timeout=0.5)" in source
+    ), "next_msg must be called with timeout=0.5"
+    # Check the except clause catches TimeoutError only
+    assert "except TimeoutError:" in source, "except clause must catch TimeoutError"
+
+
+def test_except_clause_not_redundant() -> None:
+    """The except clause should not catch both asyncio.TimeoutError and TimeoutError."""
+    source = inspect.getsource(subscriber.NATSSubscriberThread._subscribe_loop)
+    # On Python 3.10+, catching both is redundant
+    assert (
+        "except (asyncio.TimeoutError, TimeoutError):" not in source
+    ), "except clause must not catch both asyncio.TimeoutError and TimeoutError"
+    assert (
+        "except TimeoutError:" in source
+    ), "except clause should catch TimeoutError only"
--- a/tests/unit/nats/test_subscriber_polling.py
+++ b/tests/unit/nats/test_subscriber_polling.py
@@ -28,12 +28,17 @@ from hephaestus.nats.subscriber import NATSSubscriberThread
 
 
 class _FakeNatsTimeoutError(asyncio.TimeoutError):
-    """Stand-in for ``nats.errors.TimeoutError``.
+    """Stand-in for the inner timeout ``next_msg`` raises.
 
-    The real ``nats.errors.TimeoutError`` subclasses ``asyncio.TimeoutError``,
-    which is what ``next_msg`` raises on its own ``timeout``. Subclassing the
-    same base here proves the loop's ``except TimeoutError`` clause still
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
Failed to parse structured output from analysis

tests/unit/nats/test_subscriber_polling.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+import inspect
+
+from hephaestus.nats import subscriber
+
+
+def test_polling_loop_source_code_does_not_use_asyncio_wait_for() -> None:
+    """Regression: source code must not contain asyncio.wait_for around next_msg (#753)."""
+    source = inspect.getsource(subscriber.NATSSubscriberThread._subscribe_loop)
```
mvillmow: These tests assert on the *text* of the source via `inspect.getsource`, not on behavior. None of the three instantiate `NATSSubscriberThread` or run `_subscribe_loop`, so they add zero coverage of the changed runtime behavior (poll → continue on `TimeoutError`, drain, dispatch). A test that merely greps the function body passes for any code lacking the substring `asyncio.wait_for`, regardless of correctness. The plan committed to a behavioral test (drive the loop with `AsyncMock(side_effect=TimeoutError)`, assert it polls past timeouts and never calls `asyncio.wait_for`). The sibling `test_subscriber.py:375` already shows the `asyncio.run(t._subscribe_loop())` harness is available — reuse it. Replace these with a behavioral test that patches `nats.connect` and asserts the loop continues on inner timeout and exits via `_stop_event`.

tests/unit/nats/test_subscriber_polling.py
```diff
@@ -0,0 +1,37 @@
...
+    # Check the except clause catches TimeoutError only
+    assert "except TimeoutError:" in source, "except clause must catch TimeoutError"
+
+
+def test_except_clause_not_redundant() -> None:
+    """The except clause should not catch both asyncio.TimeoutError and TimeoutError."""
+    source = inspect.getsource(subscriber.NATSSubscriberThread._subscribe_loop)
+    # On Python 3.10+, catching both is redundant
+    assert (
+        "except (asyncio.TimeoutError, TimeoutError):" not in source
```
mvillmow: Hard-coding the exact string `"except (asyncio.TimeoutError, TimeoutError):"` makes this guard brittle to formatting: if a future `ruff format` pass wrapped that except across lines, the substring check would silently pass while the redundant dual-catch persisted. A behavioral test (see the major comment above) or an AST-based check of the handler's caught exception types would be robust to whitespace/reflow.

mvillmow (commented):
Production fix is correct/complete/minimal (removes dead asyncio.wait_for, narrows except). Tests are source-grep assertions with no behavioral coverage — major test-quality finding, not a correctness blocker.

tests/unit/nats/test_subscriber_polling.py
mvillmow: Addressed in 648b1c85. The three `inspect.getsource`/grep tests are gone, replaced by behavioral tests that drive the real `_subscribe_loop` via `asyncio.run` with the nats-py surface mocked (`AsyncMock(side_effect=...)`), reusing the same harness as test_subscriber.py:375. They assert the loop continues past an inner `next_msg` timeout, dispatches the next message, exits via `_stop_event`, and polls with a bounded timeout. Crucially these behavioral tests caught a *real* bug the grep tests masked: on the Python 3.10 CI runner the narrowed `except TimeoutError` let `asyncio.TimeoutError` escape and crash the loop (the two classes are distinct pre-3.11). Fixed by restoring `except (asyncio.TimeoutError, TimeoutError)`.

tests/unit/nats/test_subscriber_polling.py
mvillmow: Addressed in 648b1c85. The brittle hard-coded `except (asyncio.TimeoutError, TimeoutError):` substring assertion is removed. Robustness now comes from a behavioral parametrized test (`test_both_timeout_aliases_are_caught`) that drives `_subscribe_loop` raising each timeout alias and asserts the loop swallows it (reaches `drain`), independent of source whitespace/reflow. A future single-name narrowing of the except clause now fails this test rather than silently passing a substring grep.

PR #965, opened 2026-06-05
fix(cli): handle ragged rows in format_table

### LINKED ISSUE #774: [MINOR] cli/utils.format_table doesn't handle ragged rows
## Finding

format_table calls max(...) on all_rows after a length check, but for rows with mixed length it computes column widths from the union, which can break alignment when later rows have fewer cells.

**Location:** `hephaestus/cli/utils.py:213-217`

## Proposed fix

Pad short rows to the max column count before computing widths.

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S14

---
Tracking: #708

mvillmow:
## Summary

Fixed `format_table` to handle ragged input (rows with different column counts). Previously, column widths were computed from the union of all row lengths but the render loop short-circuited short rows without emitting padding cells, causing misalignment.

The fix normalizes all rows to the maximum column count with empty-string padding before computing widths and rendering, ensuring consistent column alignment.

## Test plan

- ✅ All 4 new ragged-row tests pass (pad-to-max-columns, widest-cell, headers-shorter-than-rows, separator-dash-spans)
- ✅ All 39 existing CLI tests pass (no regressions on format_output callers)
- ✅ Style and type checking pass

Closes #774

### DIFF
--- a/hephaestus/cli/utils.py
+++ b/hephaestus/cli/utils.py
@@ -205,26 +205,23 @@ def format_table(
     if not all_rows:
         return ""
 
-    # Calculate column widths
-    col_widths = [
-        max(len(str(row[i])) for row in all_rows if i < len(row))
-        for i in range(max(len(row) for row in all_rows))
-    ]
-
-    # Handle case where there are no columns
-    if not col_widths:
+    # Normalize ragged rows: pad short rows with "" to the max column count.
+    num_cols = max((len(row) for row in all_rows), default=0)
+    if num_cols == 0:
         return ""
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
POLICY VIOLATION: Closes (PR body empty, no 'Closes #774' line), auto-merge (must be false during review, is true), signed-commits (signing array is empty)

mvillmow (commented):
LGTM — normalize-then-render correctly fixes ragged-row alignment; all empty-input contracts preserved, 5 proportional tests cover padding/widths/header-shorter/separator/empty-inner-list. All P1–P7 pass.

PR #966, opened 2026-06-05
fix(automation): validate incompatible agent flags to prevent silent no-ops

### LINKED ISSUE #773: [MINOR] agent_stage --sandbox/--approval silently no-op for --agent=claude
## Finding

--sandbox and --approval are defined here but are silently passed through run_codex_session only — they're ignored when --agent=claude. Operators may set '--approval=on-request --agent=claude' and have the flag silently no-op.

**Location:** `hephaestus/automation/agent_stage.py:25-36`

## Proposed fix

Either (a) error when these flags are used with an agent that doesn't honor them, OR (b) emit a one-time warning in main().

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S14

---
Tracking: #708

mvillmow:
## Summary

Reject --approval and certain --sandbox values when --agent=claude is selected, as these flags are only honored by --agent=codex. Operators no longer have flag values silently ignored.

## Changes

- Added `validate_agent_flags()` to check for flag-agent incompatibilities at parse time
- `--approval` with non-default values is rejected (only applies to codex)
- `--sandbox=danger-full-access` is rejected (only applies to codex)
- `--sandbox=read-only` is explicitly allowed (it IS honored by run_claude_text)

## Test Plan

- ✅ `test_main_rejects_approval_flag_with_claude_agent` — confirms --approval rejection
- ✅ `test_main_rejects_danger_full_access_sandbox_with_claude_agent` — confirms sandbox rejection
- ✅ `test_main_allows_read_only_sandbox_with_claude_agent` — regression guard for allowed sandbox value
- ✅ `test_main_allows_default_flags_with_claude_agent` — defaults still work
- ✅ `test_main_allows_approval_with_codex_agent` — codex path unaffected
- ✅ All existing tests still pass

Closes #773

### DIFF
--- a/hephaestus/automation/agent_stage.py
+++ b/hephaestus/automation/agent_stage.py
@@ -138,6 +138,40 @@ def run_codex(
     return 0
 
 
+# Flag values that silently no-op when --agent=claude is selected.
+# - `approval` is not a parameter of run_claude_text at all, so any
+#   non-default value (i.e. != "never") is a no-op.
+# - `sandbox="read-only"` IS honored (run_claude_text:125 gates
+#   --permission-mode on it), so only `danger-full-access` is a no-op.
+# See issue #773.
+_CLAUDE_NOOP_VALUES: tuple[tuple[str, str, frozenset[str]], ...] = (
+    ("approval", "--approval", frozenset({"untrusted", "on-request"})),
+    ("sandbox", "--sandbox", frozenset({"danger-full-access"})),
+)
+
+
+def validate_agent_flags(parser: argparse.ArgumentParser, args: argparse.Namespace) -> None:
+    """Reject flag values that the selected agent does not honor.
+
+    Only fires when the operator EXPLICITLY passed ``--agent=claude``. When
+    ``--agent`` is omitted, ``resolve_agent`` will auto-detect at run time and
+    that path keeps its existing semantics. See issue #773.
+    """
+    if args.agent != "claude":
+        return
+    offending: list[str] = []
+    for attr, flag, noop_values in _CLAUDE_NOOP_VALUES:
+        value = getattr(args, attr)
+        if value in noop_values:
+            offending.append(f"{flag}={value}")
+    if offending:
+        parser.error(
+            "--agent=claude does not honor "
+            + ", ".join(offending)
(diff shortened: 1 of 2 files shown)


hephaestus/automation/agent_stage.py
```diff
@@ -138,6 +138,40 @@ def run_codex(
...
+    offending: list[str] = []
+    for attr, flag, noop_values in _CLAUDE_NOOP_VALUES:
+        value = getattr(args, attr)
+        if value in noop_values:
+            offending.append(f"{flag}={value}")
+    if offending:
+        parser.error(
+            "--agent=claude does not honor "
+            + ", ".join(offending)
+            + " (these flag values only apply to --agent=codex)"
```
mvillmow: The error message says '(these flag values only apply to --agent=codex)'. This becomes inaccurate if a third agent backend is added that also doesn't honor these flags. Prefer: '(these flag values are not supported by the claude agent)' — describes the constraint from claude's side rather than asserting codex exclusivity.

mvillmow (commented):
POLICY VIOLATION: Closes #N line absent from PR body; auto_merge_enabled=true must be false until GO; commit signing array is empty (no signed commits). Code quality is solid — one minor POLA issue in the error message wording.

mvillmow (commented):
POLICY VIOLATION: Closes (PR body empty, no Closes #773), auto-merge (enabled=true during review), signed-commits (empty array, no commits verified). Code quality is sound — three policy gates must be fixed before GO.

hephaestus/automation/agent_stage.py
mvillmow: Fixed in commit 064ec84. Changed the error message from "(these flag values only apply to --agent=codex)" to "(these flag values are not supported by the claude agent)" — now describes the constraint from claude's perspective rather than asserting codex exclusivity, which remains accurate if additional agent backends are added in the future.

mvillmow (commented):
Fixed the unresolved review thread: changed error message from '(these flag values only apply to --agent=codex)' to '(these flag values are not supported by the claude agent)' in commit 064ec84. The new wording describes the constraint from claude's perspective and remains accurate if additional agent backends are added in the future.

PR #983, opened 2026-06-05
refactor(logging): drop dead _context_extras branch in JsonFormatter

### LINKED ISSUE #795: [NITPICK] logging/formatters.py:69-78 _context_extras branch is dead in-tree
## Finding

The _context_extras branch is reached only if a caller manually sets that private attribute on the LogRecord; in the rest of the package ContextLogger.process does NOT set _context_extras, so this branch is effectively dead in-tree.

**Location:** `hephaestus/logging/formatters.py:69-78`

## Proposed fix

Delete the dead branch (YAGNI), or document the external use-case that justifies keeping it.

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S09

---
Tracking: #708

mvillmow:
## Summary

Remove the dead `_context_extras` conditional branch from `JsonFormatter.format()` that was never set by any caller in the package. This aligns with YAGNI — the branch checks for a non-existent use case and adds documentation burden.

## Implementation

- Delete the `if hasattr(record, "_context_extras"): ... else: ...` block (lines 68-78)
- Simplify to a single dict-comprehension that recovers extras via `_DEFAULT_RECORD_ATTRS` set-difference
- Add regression test ensuring `_context_extras` as a plain attribute is treated as a normal extra, not a sentinel
- All 72 existing logging + integration tests pass

## Verification

✅ Zero remaining `_context_extras` references outside of the new regression test
✅ All 12 JsonFormatter tests pass (including new regression test)
✅ Full logging test suite passes (72 tests)
✅ Integration tests pass (ContextLogger / setup_logging / correlation_id)
✅ Ruff linting and formatting pass
✅ Commit cryptographically signed

Closes #795

### DIFF
--- a/hephaestus/logging/formatters.py
+++ b/hephaestus/logging/formatters.py
@@ -66,16 +66,10 @@ class JsonFormatter(logging.Formatter):
         }
 
         # Merge extra/context fields, prefixing collisions with ``ctx_``.
-        if hasattr(record, "_context_extras"):
-            extras: dict[str, Any] = record._context_extras
-        else:
-            # Fall back to detecting manually-set extra keys by comparing
-            # against the default LogRecord attributes.
-            extras = {
-                k: v
-                for k, v in record.__dict__.items()
-                if k not in _DEFAULT_RECORD_ATTRS and k != "_context_extras"
-            }
+        # ``LoggerAdapter`` (used by ``ContextLogger``) flattens its ``extra``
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
POLICY VIOLATION: Closes #N missing (PR body empty), auto_merge_enabled=true must be false before GO, commit signing array empty (no signed commits on record).

PR #992, opened 2026-06-05
feat(automation): add audit reviewer for batch PR analysis with sub-agent dispatch

mvillmow:
## Summary

Adds `AuditReviewer` class and `hephaestus-audit-prs` CLI that reviews ALL open PRs using a coordinator+sub-agent dispatch pattern (one sub-agent per PR via the Task tool), mirroring the existing `address_review` mechanism.

Each sub-agent analyses a PR's diff, CI status, and description, then returns inline review comments posted via `gh_pr_review_post`.

## New modules

- **`hephaestus/automation/audit_reviewer.py`** — AuditReviewer class + CLI entry point
- **`hephaestus/automation/prompts/audit.py`** — Coordinator prompt (batches 10 PRs at a time, fences untrusted content)

## Modified modules

- **`github_api.py`** — Added `gh_pr_list_open()` and `_derive_ci_status()` helper
- **`session_naming.py`** — Added `AGENT_AUDIT_REVIEWER` constant
- **`prompts/__init__.py`** — Re-exports audit prompt symbols
- **`pyproject.toml`** — Added console script + coverage omit
- **`README.md` + `docs/index.md`** — Documented new console script + linked audit reviewer docs
- **`tests/integration/test_orchestration_smoke.py`** — Added audit_reviewer to OMITTED_MODULES and CONSOLE_SCRIPTS

## Documentation

- **`docs/audit-reviewer.md`** — Comprehensive documentation covering architecture, CLI usage, programmatic API, report format, coordinator prompt, error handling, environment variables, and examples

## Tests

- 13 tests for `_derive_ci_status`
- 4 tests for `gh_pr_list_open`
- 7 tests for audit coordinator prompt
- 43 tests for audit_reviewer module (unit: parsing, posting, reporting, CLI, coordinator, run orchestration, edge cases)
- 11 integration smoke tests for audit_reviewer (importability, key symbols, parsing integration, report writing, summary printing, construction)

Closes #994

PR #1001, opened 2026-06-06
fix(deps): tighten urllib3 pin to >=2.7.0 to prevent CVE regression

### LINKED ISSUE #1000: fix(deps): tighten urllib3 pin to >=2.7.0 to prevent CVE regression
## Summary

Add an explicit `urllib3 >=2.7.0` pin to `pixi.toml` to prevent the lock file from accidentally downgrading below the fix version for CVE-2026-44431 and CVE-2026-44432.

## Background

On PR #992, we removed CVE-2026-44431 and CVE-2026-44432 from the pip-audit suppression ledger after updating urllib3 from 2.6.3 to 2.7.0 via `pixi update`. Both CVEs are fixed in urllib3 >= 2.7.0.

However, urllib3 is a transitive dependency (via pygithub → requests), not explicitly pinned. The nearest explicit pin is the loose `>=1.26.0` on other transitive packages. Without an explicit pin, a future `pixi update` could silently downgrade urllib3 back below 2.7.0, reintroducing the CVEs.

## Fix

Add `urllib3 >=2.7.0` to the `[dependencies]` section of pixi.toml, following the same pattern as the existing `pyjwt >=2.13.0` security pin.

mvillmow:
Closes #1000

## Summary

Add an explicit `urllib3 >=2.7.0` pin to `pixi.toml` to prevent the lock file from accidentally downgrading below the fix version for CVE-2026-44431 and CVE-2026-44432.

## Background

On PR #992, we removed CVE-2026-44431 and CVE-2026-44432 from the pip-audit suppression ledger after updating urllib3 from 2.6.3 to 2.7.0 via `pixi update`. Both CVEs are fixed in urllib3 >= 2.7.0.

However, urllib3 is a transitive dependency (via pygithub → requests), not explicitly pinned. Without an explicit pin, a future `pixi update` could silently downgrade urllib3 back below 2.7.0, reintroducing the CVEs.

## Change

Add `urllib3 = ">=2.7.0"` to the `[dependencies]` section of pixi.toml, following the same pattern as the existing `pyjwt >=2.13.0` security pin for transitive dependencies.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -41,6 +41,11 @@ pip = "*"
 pydantic = ">=2.12.5,<3"
 pygments = ">=2.20.0"  # CVE-2026-4539
 pygithub = ">=2.9.1,<3"
+# Transitive via pygithub/requests. Pinned to prevent the lock from
+# downgrading below the fix version for CVE-2026-44431/44432 (fixed in
+# 2.7.0); without this, a future pixi update could silently reintroduce
+# the CVEs that were cleared in PR #992.
+urllib3 = ">=2.7.0"
 # Transitive via pygithub. Pinned to pull the security fix for
 # PYSEC-2026-175/177/178/179 (all fixed in 2.13.0); without this the locked
 # env resolves to 2.12.1 and pip-audit fails the dependency scan.


mvillmow:
Closing in favor of a broader audit of remaining suppressions - see issue #1000 for tracking.

PR #1009, opened 2026-06-06
refactor(automation): break circular dependency in implementer chain

### LINKED ISSUE #714: [MAJOR] Deferred-imports mask a circular dependency in implementer chain
## Background

The implementer module chain closes only because back-pointers are deferred to function bodies:

```python
# implementer.py:72-83 — top-level forward imports
from .implementer_cli import (...)
from .implementer_phase_runner import MAX_REVIEW_ITERATIONS, ImplementationPhaseRunner

# implementer_cli.py:173 — back-pointer inside function
def some_func(...):
    from . import implementer as _impl  # deferred to dodge cycle

# implementer_phase_runner.py:80, 190 — back-pointer inside function/method
def some_method(...):
    from . import implementer as _impl_mod
```

This is the bytecode-import equivalent of "we wrap it in try/except because we know it's flaky" — it works today, but:

- Any future module-level import of one of those private names creates an ImportError surface invisible to mypy
- mypy/ruff cannot analyze the import graph statically
- POLA failure: contributors expect normal Python import semantics

## Acceptance Criteria

- [ ] Resolve the cycle by extracting shared symbols into a separate module (e.g. `implementer_types.py`) that both `implementer.py` and `implementer_phase_runner.py` import top-level
- [ ] Remove all `from . import implementer as _impl*` deferred imports
- [ ] Mypy passes with `--no-implicit-reexport` on the automation subpackage

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S03 (Architecture)

---
Tracking: #708

mvillmow:
## Summary

Resolves #714 by breaking the circular dependency between `implementer.py` and its extracted children (`implementer_cli.py`, `implementer_phase_runner.py`). The cycle was previously masked by deferred imports inside function bodies.

### Changes

1. **implementer_phase_runner.py**: Added top-level imports for 9 patchable symbols previously fetched through `_impl_module` property. Deleted the property and rewrote all 7 call-site locations to use direct imports.

2. **implementer_cli.py**: Removed `main()` function (relocated to `implementer.py`) and pruned unused imports.

3. **implementer.py**: Added `main()` function definition. Removed re-export of `main` from `implementer_cli`. Preserved `_parse_args` and `_setup_logging` re-exports for backward compatibility.

4. **Tests**: Retargeted 7 patches from `implementer.*` to `implementer_phase_runner.*` where the call sites now live. Updated contract tests in `test_implementer_cli.py`. Added regression test `test_implementer_no_cycle.py` with AST-based guards.

### Verification

- ✅ All automation tests pass (1089 tests)
- ✅ Regression test verifies no deferred imports remain
- ✅ Standalone import check: both siblings import without triggering implementer.py
- ✅ Console-script entry point (`implementer:main`) continues to work
- ✅ Ruff linting passes on all modified files
- ✅ Both commits cryptographically signed

### Implementation Details

The fix uses direct source-module imports rather than creating an indirection module, following KISS and POLA principles. All call sites now use explicit module-level bindings instead of dynamic lookups.

Closes #714

### DIFF
--- a/hephaestus/automation/_review_phase.py
+++ b/hephaestus/automation/_review_phase.py
@@ -28,16 +28,18 @@ from hephaestus.agents.runtime import (
     run_codex_text,
 )
 
+from . import review_state
 from ._stage_context import StageMixin
 from .address_review import run_address_fix_session
 from .claude_invoke import (
     INFRA_ERROR_REVIEW_TEXT,
     SESSION_EXPIRED_PHRASES,
+    invoke_claude_with_session,
     parse_review_verdict,
 )
 from .claude_models import implementer_model, reviewer_model
 from .claude_timeouts import implementer_claude_timeout
-from .git_utils import issue_ref, pr_ref, run
+from .git_utils import get_repo_slug, issue_ref, pr_ref, run
 from .github_api import (
     gh_current_login,
     gh_issue_add_labels,
@@ -53,6 +55,7 @@ from .pr_manager import (
 from .pr_reviewer import gather_impl_review_context, review_pr_inline
 from .prompts import get_impl_loop_review_prompt, get_impl_resume_feedback_prompt
 from .review_validator import validate_prior_comments_addressed
+from .session_naming import AGENT_IMPLEMENTER, current_trunk_githash  # noqa: F401
 from .state_labels import STATE_SKIP
 
 if TYPE_CHECKING:
@@ -907,14 +910,13 @@ class ReviewPhase(StageMixin):
         :data:`PLAN_COMMENT_MARKER`, skipping comments whose body starts with
         :data:`review_state.PLAN_REVIEW_PREFIX`. The PLAN_REVIEW comment is
         the one whose body starts with ``review_state.PLAN_REVIEW_PREFIX``.
-        Best-effort: any fetch failure yields empty strings (looked up via
-        ``_impl_module`` so tests can patch
-        ``hephaestus.automation.implementer.review_state``).
+        Best-effort: any fetch failure yields empty strings. ``review_state``
+        is imported top-level so tests patch its internals at
+        ``hephaestus.automation._review_phase.review_state``.
         """
         plan_text = ""
         plan_review_text = ""
         try:
-            review_state = self._impl_module.review_state
             comments = review_state._fetch_issue_comments_graphql(issue_number)
             for comment in comments:
                 body = comment.get("body", "")
@@ -1027,13 +1029,12 @@ class ReviewPhase(StageMixin):
         # the Claude path — it's still consumed by the codex branch above.
         # ``recreate_on_resume_failure=False`` propagates the underlying error
         # so we can preserve the "stop iterating on expiry" contract.
-        _impl_mod = self._impl_module
-        repo_slug = _impl_mod.get_repo_slug(self.repo_root)
+        repo_slug = get_repo_slug(self.repo_root)
         try:
-            _impl_mod.invoke_claude_with_session(
+            invoke_claude_with_session(
                 repo=repo_slug,
                 issue=issue_number,
-                agent=_impl_mod.AGENT_IMPLEMENTER,
+                agent=AGENT_IMPLEMENTER,
                 prompt=prompt,
                 model=implementer_model(),
                 cwd=worktree_path,
--- a/hephaestus/automation/implementer_cli.py
+++ b/hephaestus/automation/implementer_cli.py
@@ -1,18 +1,18 @@
-"""Command-line entry point for the bulk issue implementer.
+"""Argument parsing and logging setup for the bulk issue implementer CLI.
 
-This module holds the argument-parsing, logging-setup, and ``main()`` entry
-point that were previously defined inline in :mod:`.implementer`. They were
-extracted to separate the CLI/entry-point concern from the
+This module holds the argument-parsing and logging-setup helpers that were
+previously defined inline in :mod:`.implementer`. They were extracted to
+separate the CLI plumbing from the
 :class:`~hephaestus.automation.implementer.IssueImplementer` orchestration
 class (SRP — see #468).
 
-``main`` deliberately resolves its patchable collaborators
-(``gh_list_open_issues``, ``get_repo_root``, ``CursesUI``, ``IssueImplementer``)
-through the :mod:`.implementer` module namespace rather than importing them
-directly. This preserves the existing test contract, where tests patch those
-names via ``patch.object(implementer, "...")`` and call ``implementer.main()``
-— the lookup must happen where those tests patch, not where the symbols are
-defined.
+The ``main()`` entry point lives in :mod:`.implementer` (not here): keeping it
+beside ``IssueImplementer`` lets it resolve its collaborators
+(``gh_list_open_issues``, ``get_repo_root``, ``IssueImplementer``) through that
+module's own namespace with no deferred import, which breaks the import cycle
+this module's ``main`` previously required (#714). Tests still patch those
+collaborators at ``hephaestus.automation.implementer.<name>`` and call
+``implementer.main()`` unchanged.
 """
 
 from __future__ import annotations
@@ -21,16 +21,13 @@ import argparse
 import logging
 from pathlib import Path
 
-from hephaestus.agents.runtime import add_agent_argument, resolve_agent
+from hephaestus.agents.runtime import add_agent_argument
 from hephaestus.automation._review_utils import add_max_workers_arg
 from hephaestus.cli.utils import (
     add_dry_run_arg,
     add_json_arg,
     add_version_arg,
-    emit_json_status,
 )
-
-from .models import ImplementerOptions
 
 
 def _setup_logging(verbose: bool = False, log_dir: Path | None = None) -> None:
@@ -171,83 +168,3 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         parser.error("Cannot specify both --epic and --issues")
 
     return args
-
-
-def main() -> int:
-    """Execute the issue implementation workflow.
-
-    Returns:
-        Exit code: 0 on success, 1 on failure, 130 on keyboard interrupt
-
-    """
-    # Resolve patchable collaborators through the implementer module so that
-    # existing tests patching ``implementer.gh_list_open_issues`` /
-    # ``implementer.get_repo_root`` / ``implementer.CursesUI`` /
-    # ``implementer.IssueImplementer`` continue to intercept these lookups.
-    from . import implementer as _impl
-
-    args = _parse_args()
-    agent = resolve_agent(args.agent)
-
-    state_dir = _impl.get_repo_root() / "build" / ".issue_implementer"
-    _setup_logging(args.verbose, log_dir=state_dir)
-
-    log = logging.getLogger(__name__)
-
-    # Auto-discover all open issues when neither --issues nor --epic is given
-    if not args.health_check and not args.epic and not args.issues:
-        discovered = _impl.gh_list_open_issues()
-        log.info(
-            "No --issues/--epic given; discovered %s open issues: %s", len(discovered), discovered
-        )
-        args.issues = discovered
-
-    options = ImplementerOptions(
-        epic_number=args.epic or 0,
-        issues=args.issues or [],
-        agent=agent,
-        analyze_only=args.analyze,
-        health_check=args.health_check,
-        resume=args.resume,
-        max_workers=args.max_workers,
-        skip_closed=not args.no_skip_closed,
-        auto_merge=not args.no_auto_merge,
-        dry_run=args.dry_run,
-        enable_advise=not args.no_advise,
-        enable_learn=not args.no_learn,
-        enable_follow_up=not args.no_follow_up,
-        enable_ui=not args.no_ui and not args.json,
-        include_nitpicks=args.nitpick,
-    )
-
-    if args.health_check:
-        log.info("Running health check")
-    elif args.issues:
-        log.info("Starting implementation of issues: %s", args.issues)
-    else:
-        log.info("Starting implementation of epic #%s", args.epic)
-
-    from hephaestus.utils.terminal import terminal_guard
-
-    with terminal_guard():
-        try:
-            implementer = _impl.IssueImplementer(options)
-            results = implementer.run()
-
-            if not args.health_check and not args.analyze:
-                failed = [num for num, result in results.items() if not result.success]
-                if failed:
-                    log.error("Failed to implement %s issue(s): %s", len(failed), failed)
-                    if args.json:
-                        emit_json_status(1, issues=args.issues or [], failed=failed)
-                    return 1
-
-            log.info("Complete")
-            if args.json:
-                emit_json_status(0, issues=args.issues or [], epic=args.epic or 0)
-            return 0
-        except KeyboardInterrupt:
-            log.warning("Interrupted by user")
-            if args.json:
-                emit_json_status(130, message="interrupted")
-            return 130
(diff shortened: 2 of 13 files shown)


hephaestus/automation/implementer_cli.py
```diff
@@ -1,18 +1,14 @@
...
 class (SRP — see #468).
 
-``main`` deliberately resolves its patchable collaborators
-(``gh_list_open_issues``, ``get_repo_root``, ``CursesUI``, ``IssueImplementer``)
-through the :mod:`.implementer` module namespace rather than importing them
-directly. This preserves the existing test contract, where tests patch those
-names via ``patch.object(implementer, "...")`` and call ``implementer.main()``
-— the lookup must happen where those tests patch, not where the symbols are
-defined.
+The ``main()`` entry point was relocated to :mod:`.implementer` (#714) to
```
mvillmow: The docstring says main 'is still re-exported there for backward compatibility' but main is now *defined* in implementer.py, not re-exported from implementer_cli. Change to: 'The main() entry point was relocated to :mod:`.implementer` (#714) to break a circular dependency and is now defined there directly.'

tests/unit/automation/test_implementer_no_cycle.py
```diff
@@ -0,0 +1,94 @@
...
+        if node.module is None and node.level == 1:
+            return any(a.name == "implementer" for a in node.names)
+        if node.module == "implementer" and node.level == 1:
+            return True
+        if node.module == "hephaestus.automation.implementer":
+            return True
+    return False
+
+
+def _walk_runtime(tree: ast.Module) -> None:
```
mvillmow: _walk_runtime declares `-> None` but contains `yield` statements, making it a generator. The return type must be `-> Iterator[ast.AST]` (with `from collections.abc import Iterator` or `from typing import Iterator`). As written, mypy will flag this as a type error when run with --no-implicit-reexport.

mvillmow (commented):
POLICY VIOLATION: Closes (PR description empty, no 'Closes #714' line), auto_merge (auto_merge_enabled=true, must be false during review), signed-commits (commit signing state array is empty, no verified commits)

hephaestus/automation/implementer_cli.py
mvillmow: Updated docstring to clarify main() was relocated to implementer.py, not re-exported

tests/unit/automation/test_implementer_no_cycle.py
mvillmow: Added Iterator import and fixed _walk_runtime return type from None to Iterator[ast.AST]

mvillmow (commented):
POLICY VIOLATION: Closes (PR body empty, no Closes #714 line), auto-merge (must be disabled during review), signed-commits (array empty). One code finding: current_trunk_githash missing from phase_runner import block but retargeted in tests.

hephaestus/automation/implementer_phase_runner.py
```diff
@@ -39,20 +39,29 @@
...
+# ``hephaestus.automation.implementer_phase_runner.<X>`` (#714).
+from . import review_state
+from ._review_utils import find_pr_for_issue
 from .address_review import (
     resolve_addressed_threads,
     run_address_fix_session,
 )
 from .advise_runner import run_advise
 from .claude_invoke import (
     SESSION_EXPIRED_PHRASES,
```
mvillmow: `current_trunk_githash` is retargeted in `test_implementer_loop.py:315` to `hephaestus.automation.implementer_phase_runner.current_trunk_githash`, but the import block added here (`from .session_naming import AGENT_ADVISE, AGENT_IMPLEMENTER`) does NOT include `current_trunk_githash`. The test will fail because `monkeypatch.setattr` on a name that doesn't exist in the module raises `AttributeError`. Add `current_trunk_githash` to the session_naming import: `from .session_naming import AGENT_ADVISE, AGENT_IMPLEMENTER, current_trunk_githash`.

mvillmow: Added current_trunk_githash to session_naming import with noqa directive for test patchability

mvillmow (commented):
NOGO: phase runner omits current_trunk_githash from patchable imports (autouse fixture AttributeError breaks TestResumeImplWithFeedback); 3 test patches left at old implementer.* path silently no-op and leak live gh/agent calls.

hephaestus/automation/implementer_phase_runner.py
```diff
@@ -74,14 +82,11 @@
...
-
-# NOTE: ``is_plan_review_go``, ``fetch_issue_info``,
-# ``invoke_claude_with_session``, ``get_repo_slug``, ``current_trunk_githash``,
-# and ``AGENT_IMPLEMENTER`` are deliberately NOT imported here. Existing tests
-# patch them at ``hephaestus.automation.implementer.X`` so the call sites
-# below look them up dynamically through that module via
-# :meth:`._impl_module`. This preserves the test-patch contract after the
-# #597 extraction.
+from .review_state import is_plan_review_go
+from .session_naming import (  # patchable in tests
```
mvillmow: This import block omits `current_trunk_githash`, but the retargeted autouse fixture in tests/unit/automation/test_implementer_loop.py:319-322 does `monkeypatch.setattr("hephaestus.automation.implementer_phase_runner.current_trunk_githash", ...)`. monkeypatch.setattr with a string target defaults to raising=True and raises AttributeError when the attribute does not exist, so every test in TestResumeImplWithFeedback errors. The plan specified `from .session_naming import AGENT_ADVISE, AGENT_IMPLEMENTER, current_trunk_githash`; add `current_trunk_githash` here (with `# noqa: F401` since it is patch-only, matching the pattern used in implementer.py).

tests/unit/automation/test_implementer_loop.py
```diff
@@ -313,15 +313,20 @@ class TestResumeImplWithFeedback:
 
     @pytest.fixture(autouse=True)
     def _repo_lookup(self, monkeypatch: pytest.MonkeyPatch) -> None:
-        monkeypatch.setattr("hephaestus.automation.implementer.get_repo_slug", lambda _: "TestRepo")
         monkeypatch.setattr(
-            "hephaestus.automation.implementer.current_trunk_githash", lambda _: "abc1234"
+            "hephaestus.automation.implementer_phase_runner.get_repo_slug", lambda _: "TestRepo"
+        )
+        monkeypatch.setattr(
+            "hephaestus.automation.implementer_phase_runner.current_trunk_githash",
```
mvillmow: This patch target does not exist: `implementer_phase_runner` never imports `current_trunk_githash` (verified by grep — zero occurrences in the module). monkeypatch.setattr raises AttributeError for missing attributes by default, and because this fixture is autouse, all tests in TestResumeImplWithFeedback fail at setup. Fix by adding `current_trunk_githash` to the phase runner's session_naming import (per the plan), or drop this setattr if no code path in these tests reaches it.

tests/unit/automation/test_implementer_loop.py
```diff
@@ -313,15 +313,20 @@ class TestResumeImplWithFeedback:
...
+            lambda _: "abc1234",
         )
 
     def test_resume_uses_session_id_and_prompt(
         self, implementer: IssueImplementer, tmp_path: Path
     ) -> None:
-        with patch("hephaestus.automation.implementer.invoke_claude_with_session") as mock_invoke:
+        with patch(
+            "hephaestus.automation.implementer_phase_runner.invoke_claude_with_session"
```
mvillmow: Incomplete retarget in this class: `test_resume_failure_returns_false` (line 351, unchanged) still patches `hephaestus.automation.implementer.invoke_claude_with_session`. That patch now succeeds silently (implementer.py keeps the name as a `# noqa: F401` compat re-export) but no longer intercepts the call — `_resume_impl_with_feedback` reads `implementer_phase_runner.__dict__`, so the RuntimeError side_effect never fires and the REAL invoke_claude_with_session runs in a unit test. Retarget line 351 to `hephaestus.automation.implementer_phase_runner.invoke_claude_with_session` like the three patches updated in this diff.

tests/unit/automation/test_implementer.py
```diff
@@ -351,7 +351,7 @@ def test_no_plan_path_unchanged(self, impl: IssueImplementer, tmp_path: Path) ->
             patch.object(impl, "_generate_plan") as gen_plan,
             patch.object(impl, "_save_state"),
             patch(
-                "hephaestus.automation.implementer.find_pr_for_issue",
+                "hephaestus.automation.implementer_phase_runner.find_pr_for_issue",
```
mvillmow, line 353: The `is_plan_review_go` patch four lines below (line 358, unchanged) still targets `hephaestus.automation.implementer.is_plan_review_go`. After this PR the runner's gate reads the module-level binding in `implementer_phase_runner`, so that patch is a silent no-op and the real `review_state.is_plan_review_go(1)` executes — a live `gh` call in a unit test whose `plan_review_not_go is True` assertion now depends on GitHub state. Retarget it to `hephaestus.automation.implementer_phase_runner.is_plan_review_go`, consistent with the `find_pr_for_issue` retarget on this line.

tests/unit/automation/test_implementer.py
```diff
@@ -387,7 +387,7 @@ def _record_save(state: object, *args: object, **kwargs: object) -> None:
             patch.object(impl, "_has_plan", return_value=True),
             patch.object(impl, "_save_state", side_effect=_record_save),
             patch(
-                "hephaestus.automation.implementer.find_pr_for_issue",
+                "hephaestus.automation.implementer_phase_runner.find_pr_for_issue",
```
mvillmow, line 389: Same incomplete retarget as in test_no_plan_path_unchanged: the `is_plan_review_go` patch at line 394 (unchanged) still targets `hephaestus.automation.implementer.is_plan_review_go` and is now a silent no-op — the real function runs and hits GitHub live. Retarget to `hephaestus.automation.implementer_phase_runner.is_plan_review_go`.

hephaestus/automation/implementer_phase_runner.py
```diff
@@ -74,14 +82,11 @@
...
+    AGENT_ADVISE,
+    AGENT_IMPLEMENTER,
+)
```
mvillmow, line 85: This `from .session_naming import` block omits `current_trunk_githash`. The autouse fixture added in this same PR (tests/unit/automation/test_implementer_loop.py:320) does `monkeypatch.setattr("hephaestus.automation.implementer_phase_runner.current_trunk_githash", ...)`. `monkeypatch.setattr` defaults to `raising=True`, so it raises `AttributeError` because the attribute does not exist on this module — and because the fixture is `autouse=True`, every test in `TestResumeImplWithFeedback` errors at setup. The plan's import block explicitly specified `from .session_naming import AGENT_ADVISE, AGENT_IMPLEMENTER, current_trunk_githash`. Add `current_trunk_githash` here (with `# noqa: F401`, since it is patch-only, matching the pattern in implementer.py:86).

mvillmow (commented):
NOGO: production refactor is correct, but test wiring breaks the suite — autouse fixture patches non-imported current_trunk_githash (AttributeError, whole class errors); 3 patch targets not retargeted, silently no-op and leak live gh calls.

hephaestus/automation/implementer_phase_runner.py
mvillmow: Fixed: from .session_naming import now includes current_trunk_githash in implementer_phase_runner.py. The autouse fixture's setattr resolves and TestResumeImplWithFeedback no longer errors at setup.

tests/unit/automation/test_implementer.py
mvillmow: Fixed: both find_pr_for_issue and is_plan_review_go patches are now retargeted to implementer_phase_runner.* in this test, so the gate is intercepted and no live gh call runs.

tests/unit/automation/test_implementer.py
mvillmow: Fixed: is_plan_review_go is now patched at implementer_phase_runner.is_plan_review_go here too; no silent no-op, no live GitHub call.

tests/unit/automation/test_implementer_loop.py
mvillmow: Fixed: this PR was re-implemented fresh on top of current main. ALL invoke_claude_with_session patches in TestResumeImplWithFeedback (incl. test_resume_failure_returns_false) now target implementer_phase_runner.invoke_claude_with_session, so the side_effect fires and no live call escapes.

tests/unit/automation/test_implementer_loop.py
mvillmow: Fixed: current_trunk_githash is now bound at module level in implementer_phase_runner via the .session_naming import, so monkeypatch.setattr no longer raises AttributeError. TestResumeImplWithFeedback passes.

hephaestus/automation/implementer_phase_runner.py
mvillmow: Fixed: implementer_phase_runner.py now imports current_trunk_githash from .session_naming (single-line, with noqa: F401 since it is patch-only). The autouse fixture's monkeypatch.setattr target now resolves.

hephaestus/automation/_review_phase.py
```diff
@@ -53,6 +55,7 @@
 from .pr_reviewer import gather_impl_review_context, review_pr_inline
 from .prompts import get_impl_loop_review_prompt, get_impl_resume_feedback_prompt
 from .review_validator import validate_prior_comments_addressed
+from .session_naming import AGENT_IMPLEMENTER, current_trunk_githash  # noqa: F401
```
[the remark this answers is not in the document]
mvillmow: WONT-FIX: intentional design — this is a deliberate top-level patch-surface import (marked `# noqa: F401 # patch surface — see #714 cycle-break note`). #714's whole purpose is to import these symbols top-level so tests patch them here instead of via the removed `_impl_module` indirection. Ruff passes (noqa-suppressed); the code-quality bot doesn't account for the documented rationale.

hephaestus/automation/implementer_phase_runner.py
```diff
@@ -35,22 +35,26 @@
 
 from hephaestus.agents.runtime import is_codex, run_codex_text
 
+from . import review_state  # noqa: F401  # patch surface — see #714 cycle-break note
```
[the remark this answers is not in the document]
mvillmow: WONT-FIX: intentional design — this is a deliberate top-level patch-surface import (marked `# noqa: F401 # patch surface — see #714 cycle-break note`). #714's whole purpose is to import these symbols top-level so tests patch them here instead of via the removed `_impl_module` indirection. Ruff passes (noqa-suppressed); the code-quality bot doesn't account for the documented rationale.

hephaestus/automation/implementer_phase_runner.py
```diff
@@ -61,15 +65,24 @@
...
+# modules and reached by their bare names below. Tests patch them at
+# ``hephaestus.automation.implementer_phase_runner.<X>`` — the module that owns
+# these call sites — instead of through a deferred per-runner module-pointer
+# indirection. This breaks the import cycle the old indirection created with ``implementer``
+# (#714). ``current_trunk_githash`` is patch-only (used by tests).
+from .session_naming import (  # noqa: F401
+    AGENT_ADVISE,
+    AGENT_IMPLEMENTER,
+    current_trunk_githash,
+)
```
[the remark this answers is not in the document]
mvillmow: WONT-FIX: intentional design — this is a deliberate top-level patch-surface import (marked `# noqa: F401 # patch surface — see #714 cycle-break note`). #714's whole purpose is to import these symbols top-level so tests patch them here instead of via the removed `_impl_module` indirection. Ruff passes (noqa-suppressed); the code-quality bot doesn't account for the documented rationale.

Issue #455, opened 2026-05-22
chore: upgrade Pixi to v0.69.0 and migrate pixi.lock to lockfile v7


mvillmow:
## Goal

Migrate the project to Pixi **v0.69.0** and **lockfile version 7**.

Lockfile v7 (introduced in Pixi 0.68.0, 2026-05-07) removes the source-dependency
input hashes that cause `pixi.lock` to churn on every `pixi` invocation, and locks
build dependencies for reproducible source builds.

## Background

`pixi.lock` is currently version 6. The self-referential editable path dependency
(`homericintelligence-hephaestus = { path = ".", editable = true }`) combined with
hatch-vcs dynamic versioning makes Pixi rewrite the lockfile's version line on every
run. Lockfile v7 fixes this class of churn at the format level.

Issue #<short-term> handles the immediate churn with a task-based editable install
(Option 2). This issue tracks the proper long-term fix.

## Scope

This is a coordinated change that must land in a single PR — a v7 lockfile is
unreadable by older Pixi, so CI must be upgraded in lockstep:

1. Bump the pinned Pixi version from `v0.63.2`/`v0.67.2` to `v0.69.0` in **all 8
   `setup-pixi` invocations** across 5 workflows:
   - `.github/workflows/test.yml`
   - `.github/workflows/_required.yml` (3 occurrences)
   - `.github/workflows/security.yml`
   - `.github/workflows/pre-commit.yml`
   - `.github/workflows/release.yml` (3 occurrences)
2. Upgrade local Pixi to v0.69.0 (`pixi self-update`).
3. Regenerate `pixi.lock` with the new Pixi — it becomes lockfile v7.
4. Verify CI passes end-to-end with the v7 lockfile.
5. Optionally add a `requires-pixi` floor to `pixi.toml [workspace]` so contributors
   on old Pixi get a clear error.

## References

- https://prefix.dev/blog/lock-file-v7
- Pixi 0.68.0 changelog — lockfile v7 introduction

Found by the strict 1.0-readiness audit follow-up (Pixi lockfile churn).

mvillmow:
# Implementation Plan


**Verified:** `requires-pixi` is the correct field name, placed in `[workspace]`, accepting a conda version spec (`>=0.68.0` is valid syntax). Finding #1 is now resolved with documentation evidence. Here is the revised plan.

---

# Implementation Plan — Issue #455 (Revised)

[...]

PR #1010, opened 2026-06-06
fix(automation): replace substring match with prefix-match in _has_plan

### LINKED ISSUE #715: [MAJOR] implementer_phase_runner._has_plan uses substring match, reintroducing the bug class fixed in siblings
## Background

`hephaestus/automation/implementer_phase_runner.py:611-628` `_has_plan` uses naked substring matching:

```python
for comment in comments:
    body = comment.get("body", "")
    if "Implementation Plan" in body or "## Plan" in body:
        return True
```

The same project explicitly hardened sibling functions to use prefix-only matching because substring matching caused issues #455 / #468 / #484 (reviewer treating its own prior plan-review as "the plan"):

- `planner_state.py:33-48` `_comments_contain_plan` — prefix-anchored
- `plan_reviewer.py:341-376` `_get_latest_plan` — prefix-anchored

This same-bug-different-module recurrence is the canonical pattern for "after landing a bug-pattern fix, grep sibling modules for the same pattern" from the `audit-driven-remediation` skill.

## Acceptance Criteria

- [ ] Replace substring match in `_has_plan` with the same prefix-match helper used in `planner_state._comments_contain_plan`
- [ ] Add a regression test that catches the false-positive case: a comment whose body merely quotes "## Implementation Plan" inside a Plan Review block should NOT make `_has_plan` return True
- [ ] Grep the rest of the automation subpackage for the same substring-match anti-pattern

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S04 (Source Quality)

---
Tracking: #708

mvillmow:
## Summary

Fixes the substring-match anti-pattern in `implementer_phase_runner.py` that reintroduces the bug class fixed in #455/#468/#484.

A `## 🔍 Plan Review` comment quotes the plan body (containing plan headings as substrings) and was falsely triggering as "has a plan" under the old substring matching logic.

## Changes

- Import `_comments_contain_plan` from `planner_state` (canonical prefix-anchored helper)
- Import `PLAN_COMMENT_MARKER` from `models`
- Replace `_has_plan` substring matching with delegation to `_comments_contain_plan`
- Replace `_fetch_plan_and_review` substring matching with prefix-anchored check using the same constant
- Add regression test verifying Plan Review comments are not counted as plans

Closes #715

### DIFF
--- a/tests/unit/automation/test_implementer.py
+++ b/tests/unit/automation/test_implementer.py
@@ -1014,16 +1014,28 @@ class TestHasPlanPrefixMatch:
         with patch("hephaestus.automation.implementer.get_repo_root", return_value=tmp_path):
             return IssueImplementer(options)
 
-    def test_plan_review_quoting_plan_heading_is_not_a_plan(self, impl: IssueImplementer) -> None:
-        """_has_plan does not match substring in a Plan Review comment.
+    @staticmethod
+    def _mock_gh_view(comments: list[dict[str, Any]]) -> Any:
+        def mock_gh_view(*args: Any, **kwargs: Any) -> MagicMock:
+            result = MagicMock()
+            result.stdout = json.dumps({"comments": comments}).encode("utf-8")
+            return result
 
-        Regression for #455/#468/#484: a ``## 🔍 Plan Review`` body contains
-        ``## Implementation Plan`` as substring when it quotes the plan, and
+        return mock_gh_view
+
+    def test_plan_review_quoting_plan_heading_is_not_a_plan(self, impl: IssueImplementer) -> None:
+        """_has_plan does not match a Plan Review comment that merely quotes the plan.
+
+        Regression for #455/#468/#484/#715: a ``## 🔍 Plan Review`` body contains
+        ``# Implementation Plan`` as substring when it quotes the plan, and
         substring matching caused the false positive. Prefix-matching (only at
         the start of the body) must be used instead.
+
+        The fixture intentionally contains ONLY the Plan-Review comment (no
+        genuine plan), so the OLD substring code would return ``True`` here —
+        this test fails against the pre-fix code and guards the bug.
         """
         comments = [
-            {"body": "# Implementation Plan\n\nStep 1: Do something"},
             {
                 "body": (
                     "## 🔍 Plan Review\n\n"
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
POLICY VIOLATION: Closes, auto-merge, signed-commits — PR description is empty (no 'Closes #715' line); auto_merge_enabled=true must be false during implementation review; commit signing state array is empty (no signed commits recorded).

mvillmow (commented):
Failed to parse structured output from analysis

mvillmow (commented):
Failed to parse structured output from analysis

tests/unit/automation/test_implementer.py
```diff
@@ -571,3 +573,64 @@ def test_run_with_epic_but_no_issues_still_proceeds(self, tmp_path: Path) -> Non
...
+            result = MagicMock()
+            result.stdout = json.dumps({"comments": comments}).encode("utf-8")
+            return result
+
+        with patch(
+            "hephaestus.automation.implementer_phase_runner.run",
+            side_effect=mock_gh_view,
+        ):
+            # The first comment is the plan; the second quotes it but is NOT the plan
+            assert impl.phase_runner._has_plan(1) is True
```
mvillmow, line 1068: This regression test does not guard the #715 bug and passes against the OLD substring code, so it provides no protection. The `comments` fixture (line 615) includes a genuine plan comment (`# Implementation Plan...`) as the first entry, so `_has_plan` returns True for the right input — but the pre-fix code (`"Implementation Plan" in body`) also returns True on that same first comment. Revert the production change and this test still passes.

AC2 requires testing that a Plan Review which merely *quotes* the plan does NOT count. The method name (`..._is_not_a_plan`) and docstring describe exactly that False case, but the assertion is `is True` (and the inline comment says "The first comment is the plan") — the test contradicts its own name.

Fix: drop the genuine plan comment from the fixture so it contains ONLY a Plan-Review comment that quotes the plan heading, and assert `impl.phase_runner._has_plan(1) is False`. Also restore the planned positive-path test (`test_real_plan_comment_is_detected`: a body starting with `# Implementation Plan` → `is True`) so both branches of the prefix check are covered.

mvillmow (commented):
Production fix is correct (both callsites prefix-anchored, AC1/AC3 met), but the sole new regression test asserts True with a real plan present — passes against the buggy code, so AC2 coverage is absent.

tests/unit/automation/test_implementer.py
mvillmow: Addressed. The test suite now guards #715 directly:

- `test_plan_review_quoting_plan_heading_is_not_a_plan` uses a fixture containing **only** a `## 🔍 Plan Review` comment (body starts with the Plan-Review prefix, with `# Implementation Plan` appearing mid-text as a quote) and asserts `_has_plan(1) is False`. The pre-fix substring code (`"Implementation Plan" in body`) returns `True` on this fixture, so the test **fails against the old code** and genuinely guards the regression.
- `test_real_plan_comment_is_detected` restores the positive branch: a body starting with `# Implementation Plan` → `_has_plan(1) is True`.

Verified empirically: old substring logic on the Plan-Review-only fixture yields `True`, but the test asserts `False`. Both branches of the prefix check are now covered. The misleading earlier fixture (genuine plan as first comment + contradicting `is True`/name) has been removed.

PR #1054, opened 2026-06-06
feat(github): add --ascii flag to fleet_sync for ASCII-only output

### LINKED ISSUE #809: [NITPICK] fleet_sync.py:619-621 hard-coded unicode summary banner
## Finding

Hard-coded summary banner uses literal Unicode '═'/'✓' characters with no --no-color/--ascii toggle — fine in modern terminals, mildly surprising in pipelines that route stdout through ASCII-only consumers.

**Location:** `hephaestus/github/fleet_sync.py:619-621`

## Proposed fix

Add --ascii flag that swaps Unicode for ASCII equivalents.

## Discovered via

`/hephaestus:repo-analyze-strict-full` 2026-05-29, S04

---
Tracking: #708

mvillmow:
## Summary

Add `--ascii` flag to `hephaestus-fleet-sync` that swaps Unicode glyphs (══, ✓, →, —) for ASCII equivalents (==, *, ->, --) in log output. This enables safe pipeline consumption by ASCII-only consumers like CI log shippers and legacy syslog forwarders.

## Design

Uses a frozen `Symbols` dataclass with `UNICODE_SYMBOLS` and `ASCII_SYMBOLS` presets. The selected preset is passed as a keyword argument through `process_repo` → helper functions, maintaining backward compatibility and determinism without global state.

- Added `Symbols` dataclass with four glyph fields (banner, check, arrow, dash)
- Created `UNICODE_SYMBOLS` (default) and `ASCII_SYMBOLS` presets
- Updated function signatures: `process_repo`, `rebase_and_resign`, `resolve_conflict_with_agent`
- Updated all logger.info/warning calls in user-facing output to use symbols
- Added comprehensive unit tests with `TestAsciiFlag` class

## Test Plan

- All existing fleet_sync tests pass (52 total)
- New tests verify:
  - Symbols dataclass is frozen
  - Presets have expected glyphs
  - process_repo emits correct banners under both presets
- No regressions in existing functionality
- `--ascii` flag appears in help text
- No Unicode literals remain in logger calls

Closes #809

### DIFF
--- a/hephaestus/github/fleet_sync.py
+++ b/hephaestus/github/fleet_sync.py
@@ -868,7 +868,7 @@ def main() -> int:
         "--ascii",
         action="store_true",
         help="Use ASCII equivalents for Unicode glyphs in log output "
-             "(== for ══, * for ✓, -> for →, -- for —). "
+             "(== for ==, * for *, -> for ->, -- for --). "
              "Use this when piping stdout to ASCII-only consumers.",
     )
     add_json_arg(parser)
--- a/hephaestus/github/fleet_sync.py
+++ b/hephaestus/github/fleet_sync.py
@@ -43,6 +43,20 @@ from hephaestus.utils.helpers import METADATA_TIMEOUT, NETWORK_TIMEOUT
 
 logger = get_logger(__name__)
 
+
+@dataclass(frozen=True)
+class Symbols:
+    """Glyphs used in user-facing log output. Frozen for safe sharing across calls."""
+
+    banner: str
+    check: str
+    arrow: str
+    dash: str
+
+
+UNICODE_SYMBOLS = Symbols(banner="══", check="✓", arrow="→", dash="—")
+ASCII_SYMBOLS = Symbols(banner="==", check="*", arrow="->", dash="--")
+
 ORG = "HomericIntelligence"
 
 
@@ -519,7 +533,13 @@ def merge_pr(pr: PRInfo, dry_run: bool = False) -> bool:
         return False
 
 
-def rebase_and_resign(pr: PRInfo, repo_clone: Path, dry_run: bool = False) -> bool:
+def rebase_and_resign(
+    pr: PRInfo,
+    repo_clone: Path,
+    dry_run: bool = False,
+    *,
+    symbols: Symbols = UNICODE_SYMBOLS,
+) -> bool:
     """Fetch PR branch, rebase it on origin/base, re-sign all commits, push.
 
     Operates in a per-PR worktree off the shared ``repo_clone`` (#1044) rather
@@ -540,12 +560,14 @@ def rebase_and_resign(pr: PRInfo, repo_clone: Path, dry_run: bool = False) -> bo
         )
 
         if result.returncode != 0:
-            logger.warning("  Rebase failed for PR #%d — conflict detected", pr.number)
+            logger.warning(
+                "  Rebase failed for PR #%d %s conflict detected", pr.number, symbols.dash
+            )
             _git(["rebase", "--abort"], cwd=work, dry_run=dry_run, check=False)
             return False
 
         _git(["push", "--force-with-lease", "origin", branch], cwd=work, dry_run=dry_run)
-        logger.info("  ✓ Rebased and re-signed PR #%d", pr.number)
+        logger.info("  %s Rebased and re-signed PR #%d", symbols.check, pr.number)
         return True
 
     except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
@@ -597,6 +619,8 @@ def resolve_conflict_with_agent(
     repo_clone: Path,
     dry_run: bool = False,
     agent: str = "claude",
+    *,
+    symbols: Symbols = UNICODE_SYMBOLS,
 ) -> bool:
     """Spawn the selected agent to semantically resolve merge conflicts, then re-sign.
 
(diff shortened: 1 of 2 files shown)


hephaestus/github/fleet_sync.py
```diff
@@ -825,6 +852,13 @@ def main() -> int:
     )
     add_agent_argument(parser)
     parser.add_argument("--verbose", "-v", action="store_true", help="Debug logging")
+    parser.add_argument(
+        "--ascii",
+        action="store_true",
+        help="Use ASCII equivalents for Unicode glyphs in log output "
```
mvillmow: The `--help` text contains the Unicode glyphs (`══`, `✓`, `→`, `—`) that the `--ascii` flag is designed to avoid. A user in an ASCII-only terminal who runs `--help` to discover this flag will see mojibake before they can use it. Either strip Unicode from the help string, or add a note acknowledging this as a known limitation.

tests/unit/github/test_fleet_sync.py
```diff
@@ -793,3 +793,108 @@ def test_clean_success_is_ready(self, monkeypatch) -> None:
...
+        """process_repo logs ASCII banner under ASCII_SYMBOLS — no module state."""
+        monkeypatch.setattr(fleet_sync_module, "list_prs", lambda _repo: [])
+
+        import argparse
+        import logging
+
+        # Capture logger output directly
+        logged_messages = []
+
+        class TestHandler(logging.Handler):
```
mvillmow: The `TestHandler` + `underlying_logger` capture block (lines 831-840) is duplicated verbatim in `test_process_repo_emits_unicode_banner_by_default` (lines 876-886). Extract this into a `@pytest.fixture` or a private `_capture_log_messages(module)` helper to avoid the DRY violation — the pattern is 10 lines and will need to be kept in sync if `ContextLogger` internals change.

mvillmow (commented):
POLICY VIOLATION: Closes (PR body is empty — no `Closes #809` line), auto-merge (armed before implementation-GO), signed-commits (empty array — no signed commits found)

hephaestus/github/fleet_sync.py
mvillmow: Fixed help text to use ASCII equivalents (== for ==, * for *, -> for ->, -- for --)

tests/unit/github/test_fleet_sync.py
mvillmow: Extracted TestHandler capture pattern into @pytest.fixture `capture_fleet_sync_logs()`; refactored both ASCII banner tests to use it, eliminating 30+ lines of duplication

mvillmow (commented):
POLICY VIOLATION: Closes (PR description is empty — no 'Closes #809' line), auto-merge (must be false during review), signed-commits (commit signing array is empty)

hephaestus/github/fleet_sync.py
```diff
@@ -825,6 +852,13 @@ def main() -> int:
...
+             "(== for ==, * for *, -> for ->, -- for --). "
```
mvillmow: The help text is self-contradictory: it promises "ASCII equivalents for Unicode glyphs" but the parenthetical maps each ASCII glyph to itself — `== for ==`, `* for *`, `-> for ->`, `-- for --`. This conveys nothing about what `--ascii` swaps, and `--help` is the only discoverability surface for the feature. This looks like an over-correction of the plan-review concern that Unicode in the help text mojibakes under ASCII-only locales — but the fix was to drop the Unicode side, not garble the mapping. Use the team-KB ASCII-safe phrasing that both avoids mojibake AND describes the mapping, e.g.: `help="Use ASCII fallbacks (==, *, ->, --) instead of Unicode box/check/arrow/dash glyphs in log output; use when piping stdout to ASCII-only consumers."` (add `# noqa: E501` if it exceeds the line limit).

tests/unit/github/test_fleet_sync.py
```diff
@@ -793,3 +834,76 @@ def test_clean_success_is_ready(self, monkeypatch) -> None:
             self._classify(monkeypatch, mergeable="MERGEABLE", state="CLEAN", ci="SUCCESS")
             == PRStatus.READY
         )
+
+
```
mvillmow, line 838: The plan and team learnings called for a test asserting `--ascii` is registered on the production parser (via an extracted `_build_parser()` or by inspecting `main()`'s parser), but `TestAsciiFlag` has no such test. The two behavioral tests pass `symbols=` directly and never go through argparse, so if the `add_argument("--ascii", ...)` call were removed from `main()`, no test would fail. Add a parser-registration assertion to close that gap — e.g. inspect the real parser's actions for an `--ascii` option whose `default is False`.

mvillmow (commented):
Design correct & plan-aligned, but --ascii --help text reads '== for ==, * for *, ...' — a meaningless ASCII-for-ASCII mapping. NOGO: fix the help text.

hephaestus/github/fleet_sync.py
mvillmow: Fixed in dd26538. Rewrote the `--ascii` help text to describe the actual Unicode→ASCII swap instead of the self-mapping: "Use ASCII fallbacks (==, *, ->, --) instead of Unicode box/check/arrow/dash glyphs in log output; use when piping stdout to ASCII-only consumers." Added a regression test (`test_ascii_help_describes_glyph_swap_not_identity_mapping`) asserting the help names Unicode and no longer contains the garbled `== for ==` mapping.

tests/unit/github/test_fleet_sync.py
mvillmow: Fixed in dd26538. Extracted `_build_parser()` from `main()` so the production parser is inspectable, and added `test_ascii_flag_registered_on_production_parser` which asserts the real parser registers `--ascii` as a store_true action with `default is False`, and that `parse_args(["--ascii"]).ascii is True`. Removing the `add_argument` call would now fail this test.

PR #1097, opened 2026-06-08
fix(automation): catch TimeoutExpired and OSError in failing-PR discovery

### LINKED ISSUE #1096: Follow-up from #819: 1 item(s) (safety)
_Consolidated follow-up from implementation of #819._

### _discover_failing_prs lets TimeoutExpired/OSError abort the drive

ci_driver.py:475 catches only (subprocess.CalledProcessError, json.JSONDecodeError). A gh timeout (subprocess.TimeoutExpired) or missing-binary (OSError) propagates uncaught to the worker-thread exception handler, marking the entire work item failed — violating the docstring contract at ci_driver.py:450: 'Empty dict on any lookup failure — discovery must never abort the drive.' Fix: widen the except clause to (subprocess.CalledProcessError, subprocess.TimeoutExpired, OSError, json.JSONDecodeError). loop_runner._count_failing_prs already catches these at loop_runner.py:650.

---

mvillmow:
## Summary

Widen the except clause in CIDriver._discover_failing_prs to catch subprocess.TimeoutExpired (gh CLI timeout) and OSError (missing/unexecutable gh binary). These transient failures should return the documented empty-dict fallback instead of escaping to the worker-thread exception handler.

Closes #1096

### DIFF
--- a/hephaestus/automation/ci_driver.py
+++ b/hephaestus/automation/ci_driver.py
@@ -472,7 +472,12 @@ class CIDriver:
                 ],
             )
             pulls: list[dict[str, Any]] = json.loads(result.stdout or "[]")
-        except (subprocess.CalledProcessError, json.JSONDecodeError) as exc:
+        except (
+            subprocess.CalledProcessError,
+            subprocess.TimeoutExpired,
+            OSError,
+            json.JSONDecodeError,
+        ) as exc:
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
POLICY VIOLATION: Closes, auto-merge-armed, signed-commits — PR body is empty (no Closes #1096), auto-merge is already enabled, and the commits signing array is empty (no signed commits confirmed).

PR #1233, opened 2026-06-12
docs(migration): sync version header to v0.9.5 + add drift guard

### LINKED ISSUE #1208: [audit][MINOR] MIGRATION.md version header is stale relative to tags
**Location:** `docs/MIGRATION.md:1-11`

Header states 'the latest released version is 0.9.2' (as of 2026-05-28), but git tags show v0.9.3, v0.9.4, v0.9.5 have since shipped. ROADMAP.md 'Last updated: 2026-05-28' similarly trails three patch releases. Not wrong policy, but the dated currency claims are out of sync with the release line — exactly the 'documentation currency' the project flags in its own ROADMAP audit-remediation list (ROADMAP.md:11).

---

mvillmow:
## Summary

`docs/MIGRATION.md` and `docs/ROADMAP.md` dated currency claims trailed the release line by three patch versions — the docs said `0.9.2` / `2026-05-28`, but the latest git tag is `v0.9.5`. This is the "documentation currency" debt the project flags in its own ROADMAP audit-remediation list.

### Changes

1. **Sync both docs** to `v0.9.5` / `2026-06-12` (`docs/MIGRATION.md:3`, `docs/ROADMAP.md:55`). `ROADMAP.md:9` is left untouched — "the strict 2026-05-28 audit" is a historical event name, not a currency claim.
2. **Add a CI-effective drift-detection regression test** (`tests/unit/docs/test_version_currency.py`) that ties the MIGRATION.md "latest released version" line to the canonical hatch-vcs authority (latest `vX.Y.Z` git tag via `_version_from_git_tag`) so the line can never silently trail the tag again.
3. **Make the guard load-bearing in CI**: `.github/workflows/test.yml` checkout now sets `fetch-depth: 0` + `fetch-tags: true` (mirroring `auto-tag.yml` / `release.yml`) so the test resolves a tag in CI instead of `None`. If tags are ever absent, the test **fails loud** with remediation guidance rather than skipping.

### Semantics

Uses `>=` (not `==`) comparison so a freshly-pushed newer tag does not instantly red `main` — it only fails for the genuinely-stale case this issue targets (`0.9.2 < 0.9.5`).

### Verification

- RED→GREEN confirmed: test fails (assertion, not skip) on the stale `0.9.2` state, passes after the sync. `0 skipped`.
- `tests/unit/version` + `tests/unit/docs`: 86 passed.
- Corpus re-grep: no stale `0.9.2` / `2026-05-28` strings remain except the historical `ROADMAP.md:9` audit name.
- pre-commit (ruff, mypy, yamllint, test-structure) passes on all changed files.

Closes #1208

### DIFF
--- a/tests/unit/docs/test_version_currency.py
+++ b/tests/unit/docs/test_version_currency.py
@@ -1,0 +1,56 @@
+"""Regression test: MIGRATION.md version claim must not trail the latest git tag.
+
+Pattern D (stop maintaining the number by hand): instead of relying on a human
+to notice the dated 'latest released version' line has gone stale, this test
+fails CI when the documented version is *older* than the canonical hatch-vcs
+version (latest vX.Y.Z git tag). See issue #1208.
+
+This test is designed to RUN in CI, not skip: the unit-test job checks out with
+``fetch-tags: true`` (see .github/workflows/test.yml). If tags are somehow
+absent, the test FAILS LOUD with remediation guidance rather than skipping --
+a guard that silently skips is not a guard.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+from hephaestus.version.consistency import _version_from_git_tag
+from hephaestus.version.parsing import parse_version_tuple
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+MIGRATION_MD = REPO_ROOT / "docs" / "MIGRATION.md"
+
+# Matches: "The latest released version is **0.9.5**"
+_LATEST_RE = re.compile(r"latest released version is \*\*(\d+\.\d+\.\d+)\*\*", re.IGNORECASE)
+
+
+def test_migration_md_version_does_not_trail_latest_git_tag() -> None:
+    """Fail if MIGRATION.md's 'latest released version' trails the newest git tag."""
+    canonical = _version_from_git_tag(REPO_ROOT)
+    if canonical is None:
+        pytest.fail(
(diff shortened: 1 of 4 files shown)


tests/unit/docs/test_version_currency.py
```diff
@@ -0,0 +1,56 @@
...
+from pathlib import Path
+
+import pytest
+
+from hephaestus.version.consistency import _version_from_git_tag
+from hephaestus.version.parsing import parse_version_tuple
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+MIGRATION_MD = REPO_ROOT / "docs" / "MIGRATION.md"
+
```
mvillmow, line 24: The regex couples this test to the exact MIGRATION.md phrasing `latest released version is **X.Y.Z**`. This is mitigated well by the `assert match is not None` below (a phrasing change fails loud with a clear message rather than silently passing), so it's acceptable as-is. Flagging only so a future editor who rewords that sentence knows to update this regex in lockstep — no change required now.

mvillmow (commented):
GO. Doc sync to v0.9.5 correct; ROADMAP:9 historical name correctly preserved. Drift guard runs in CI (tags fetched, hard-fail not skip), correct >= comparison, verified imports/placement. One minor regex-coupling note.

mvillmow (commented):
Doc version sync to v0.9.5 is correct and complete; drift test is load-bearing in CI (fetch-tags + hard-fail), uses non-racy >= semantics, reuses existing helpers. LGTM.

PR #1246, opened 2026-06-13
fix(utils): make max_delay a hard ceiling in retry backoff

### LINKED ISSUE #1206: [audit][MINOR] Jitter can exceed the documented max_delay cap
**Location:** `hephaestus/utils/retry.py:73-78`

max_delay is applied BEFORE jitter, so with jitter=True the actual sleep can reach max_delay*1.25 (acknowledged in the docstring L66-69). The cap is therefore advisory rather than a hard ceiling; a caller setting max_delay to bound worst-case latency does not actually get that bound. Applying min(delay, max_delay) after jitter (or clamping) would honor the cap's intent.

---

mvillmow:
## Summary

`_compute_backoff_delay` in `hephaestus/utils/retry.py` applied the `max_delay`
cap *before* jitter, so with `jitter=True` the actual sleep could reach
`max_delay * 1.25` — the docstring even acknowledged this. The cap was therefore
advisory rather than a hard ceiling: a caller setting `max_delay` to bound
worst-case latency did not actually get that bound.

## Fix

Re-apply `min(delay, max_delay)` *after* jitter so `max_delay` becomes a true
hard ceiling. The pre-jitter clamp is intentionally retained — it keeps jitter
proportional to the capped delay rather than to an unbounded exponential, so
high-attempt delays still spread around the cap instead of collapsing to exactly
`max_delay` (which would defeat the thundering-herd purpose of jitter). The
`max(0.1, delay)` floor remains the final operation.

## Tests

- Added `test_max_delay_hard_ceiling_with_jitter`: forces `random.uniform` to its
  `+25%` maximum and asserts the sleep never exceeds `max_delay` with
  `jitter=True` (fails RED against the old code: `2.5 <= 2.0`).
- Tightened `test_max_delay_respected`, which encoded the buggy `* 1.25`
  tolerance, to assert `<= max_delay`.
- `test_max_delay_caps_backoff` (`jitter=False`) is unchanged and remains the
  regression anchor.

All 23 tests in `test_retry.py` pass; `ruff format`, `ruff check`, and `mypy`
are clean.

Closes #1206

### DIFF
--- a/hephaestus/utils/retry.py
+++ b/hephaestus/utils/retry.py
@@ -62,12 +62,14 @@ def _compute_backoff_delay(
         attempt: Zero-based attempt index (0 for the first retry).
         initial_delay: Initial delay in seconds before first retry.
         backoff_factor: Multiplier for delay between retries.
-        max_delay: Optional cap (applied *before* jitter so the actual sleep
-            value can reach max_delay * 1.25 at most when jitter=True).
+        max_delay: Optional hard cap on the sleep value. Applied *after* jitter
+            so the returned delay never exceeds max_delay (subject to the 0.1s
+            minimum floor below).
         jitter: If True, perturb the delay by ±25 %.
 
     Returns:
-        Sleep duration in seconds (always >= 0.1).
+        Sleep duration in seconds (always >= 0.1, and <= max_delay when a cap
+        is set and max_delay >= 0.1).
 
     """
     delay: float = initial_delay * (backoff_factor**attempt)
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
LGTM — post-jitter re-clamp makes max_delay a true hard ceiling; RED-first test forces +25% jitter and asserts <=cap; bug-encoding assertion corrected; docstrings now match behavior. Minimal, correct, no findings.

PR #1247, opened 2026-06-13
feat(forensics): executable handler-ran contract guard (--verify)

### LINKED ISSUE #1207: [audit][MINOR] Coredump handler reliability relies on an unverified convention
**Location:** `hephaestus/forensics/coredump_handler.py:37-40,132-149`

Because the kernel ignores a pipe handler's exit code, all failures are logged to handler.log and the documented contract is 'a missing handler.log means the handler never ran.' This is sound, but there is no automated downstream check that enforces it; a CI artifact step that forgets to assert handler.log presence would silently lose the only failure signal. The convention is correct but un-guarded.

---

mvillmow:
## Summary

Issue #1207 observed that the coredump handler's failure-signal contract — *"a missing `handler.log` means the handler never ran"* — was documented but **un-guarded**: nothing asserted it, so a downstream CI artifact step that forgets to check `handler.log` would silently lose the only failure signal.

This adds a small, tested, reusable `verify_crash_bundle()` plus a `--verify` mode on the existing `hephaestus-coredump-handler` console script, turning the prose convention into an executable, blocking check that CI or any consumer can call.

## What changed

- **`verify_crash_bundle(log_dir) -> (verdict, detail)`** classifies a crash bundle into `OK` / `RAN_WITH_ERRORS` / `NOT_RUN`. A `wrote ` line is the authoritative success signal (so a successful capture that also logged a chmod/`COREDUMP_MAX_BYTES` WARNING is still `OK`); a missing/empty/unreadable `handler.log` is `NOT_RUN` — the silent-loss case.
- **`--verify` CLI mode** exits `0` when the handler provably ran and `3` (`VERIFY_SIGNAL_LOST_EXIT`) on `NOT_RUN`. Exit `3` is deliberately distinct from argparse's usage-error `2` and `gdb_runner`'s `2`, so a CI gate can tell "signal lost" apart from "bad command line".
- **Read-only resolution**: `--verify` never `mkdir()`s the target dir, so it cannot fabricate the directory whose absence is the signal.
- **Capture path stays loud**: positionals became optional only so `--verify` can omit the kernel tokens; a real capture invocation missing them now fails with exit `1` instead of silently no-opping.
- **Doc/shell sync**: module docstring and `scripts/shell/coredump-host-handler.sh` comment now reference the executable guard.

## Tests

`TestVerifyCrashBundle`, `TestMainVerify`, and `TestMainCaptureGuards` cover all classification branches, the exit-`3` contract, the no-fabrication invariant, the JSON envelope (`status=="error"`, `exit_code==3`, `verdict=="NOT_RUN"`), and a full-capture regression. Full forensics + validation suites: 764 passed. `ruff`, `ruff format`, and `mypy` all clean.

Closes #1207

### DIFF
--- a/hephaestus/forensics/coredump_handler.py
+++ b/hephaestus/forensics/coredump_handler.py
@@ -208,7 +208,14 @@ def verify_crash_bundle(log_dir: Path) -> tuple[str, str]:
 
     # A `wrote ` line is the authoritative success signal; a successful capture
     # may also carry a chmod/max-bytes WARNING, so `wrote ` wins over WARNING.
-    if any(" wrote " in ln or ln.startswith("wrote ") for ln in lines):
+    # Each kept line is "<iso-timestamp> <message>"; split off the timestamp and
+    # check that the message portion starts with "wrote " so that an ERROR: line
+    # whose path happens to contain the substring " wrote " is not misclassified.
+    def _message_is_wrote(ln: str) -> bool:
+        parts = ln.split(maxsplit=1)
+        return len(parts) == 2 and parts[1].startswith("wrote ")
+
+    if any(_message_is_wrote(ln) for ln in lines):
         return BUNDLE_OK, f"{log_path} records a successful capture"
     return BUNDLE_RAN_WITH_ERRORS, (
         f"{log_path} records handler activity but no successful capture "
--- a/hephaestus/forensics/coredump_handler.py
+++ b/hephaestus/forensics/coredump_handler.py
@@ -37,7 +37,10 @@ maps to), pass it as a literal ``--target-dir`` argument in the
 Exit-code note: the kernel ignores this handler's exit code entirely. Every
 failure path therefore logs to ``handler.log`` next to the core directory
 instead of relying on exit status — a missing ``handler.log`` is the only
-signal that the handler did not run at all.
+signal that the handler did not run at all. This contract is enforced by
+:func:`verify_crash_bundle` (also exposed as ``hephaestus-coredump-handler
+--verify``, which exits 3 on a lost signal); a CI artifact step SHOULD run it
+and fail the job when it reports ``NOT_RUN``.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


hephaestus/forensics/coredump_handler.py
```diff
@@ -149,6 +152,70 @@ def _log(log_dir: Path, message: str) -> None:
...
+    except OSError as exc:
+        return BUNDLE_NOT_RUN, f"{log_path} is unreadable ({exc})"
+
+    lines = [ln for ln in contents.splitlines() if ln.strip()]
+    if not lines:
+        return BUNDLE_NOT_RUN, f"{log_path} is empty — no handler activity recorded"
+
+    # A `wrote ` line is the authoritative success signal; a successful capture
+    # may also carry a chmod/max-bytes WARNING, so `wrote ` wins over WARNING.
+    if any(" wrote " in ln or ln.startswith("wrote ") for ln in lines):
```
mvillmow: The success match `" wrote " in ln` substring-scans every line, including `ERROR:` lines. Since `_log` writes `wrote {out} …` where `out` derives from `core.{pid}.{exe}.{...}` and `%e` (exe basename) can contain spaces, an `ERROR: failed to write core to …/<name with the literal ' wrote ' in it>` line could be misclassified as BUNDLE_OK. Narrow and theoretical, but you could tighten it by anchoring on the actual log line prefix (a line that *starts* with the timestamp+`wrote `) rather than a free substring scan. The `startswith("wrote ")` branch is also currently unreachable given the real line shape (`<iso-timestamp> wrote …` always contains `' wrote '`).

mvillmow (commented):
Faithful, source-verified implementation of the executable handler-ran guard; exit code 3 collision-free, JSON envelope correct, no-mkdir invariant tested. ≤2 minor nits.

mvillmow (commented):
LGTM — executable handler-ran guard correct and complete; prior substring-match minor resolved via prefix-anchored classifier with regression test; exit code 3 collision-free; JSON envelope and no-mkdir invariant verified.

Issue #1214, opened 2026-06-12
[audit][MINOR] build/ working directory is a junk drawer of automation logs


mvillmow:
**Location:** `/home/mvillmow/Projects/ProjectHephaestus/build/`

build/ contains ~100 loop_*.log files, chunk_*.json, open_prs.json, work reports, and a nested ProjectMnemosyne checkout — not packaging output. It is harmless to the wheel (sdist only-include excludes it, pyproject.toml:158-165) but pollutes the repo named 'build', risks confusing contributors expecting build artifacts, and is one stray glob away from being swept into a distribution. POLA/modularity nitpick edging into minor because of the packaging-namespace collision.

---
_Filed by strict full-coverage repo audit (2026-06-12)._

mvillmow:
# Implementation Plan


Confirmed. `S603` absent from select → `# noqa: S603` is RUF100. The conftest auto-discovers any new `scripts/*.py`, so my script is automatically subjected to the `--help` smoke test. Both fixes: add the `--help`/`-h` branch (print `__doc__`, return 0) and drop the `# noqa: S603`. I'll also add a unit test asserting the `--help` contract directly so TDD drives the required interface. Now I have everything to write the revised plan.

# Implementation Plan

## Objective

Issue #1214 flags `build/` as a "junk drawer" of automation logs (81 `loop_*.log` files, `chunk_*.json`, `audit_*.json`, `.worktrees/`, a nested `ProjectMnemosyne` clone) whose name collides with the Python packaging-output convention. Verification confirms the junk is **untracked-but-on-disk** (`git ls-files build/` → 0 results; `.gitignore:5` already ignores `build/`) and **harmless to distributions** (`pyproject.toml:158-165` `only-include` is an exact-path sdist allowlist that excludes `build/`). It is the **sanctioned** scratch location — `loop_runner.py:351` writes work reports there and `CLAUDE.md:455` mandates "all temporary files are in the build/ directory." So this is a POLA/naming-collision nit, not a misplacement. The durable fix is a **regression guard** (pre-commit hook + script) that hard-fails if any file is ever tracked under `build/`, closing the "one stray glob away from a distribution" risk structurally, plus a `CONTRIBUTING.md` note making the scratch convention explicit. No on-disk junk is deleted in this PR (a live automation loop regenerates it; deleting mid-run is a known anti-pattern).

## Approach

[...]

PR #1252, opened 2026-06-13
[CI] Machine-enforce dependency license compatibility

### LINKED ISSUE #1219: [audit][MINOR] License compatibility is asserted in NOTICE but not machine-enforced
**Location:** `/home/mvillmow/Projects/ProjectHephaestus/.github/workflows/security.yml:1-120`

NOTICE makes correct manual license-compatibility determinations (BSD/MIT/Apache/LGPL-as-optional), but no CI job (e.g. pip-licenses / reuse / a license allowlist) verifies that newly added or transitively-upgraded dependencies stay license-compatible. pip-audit (security.yml:39) checks CVEs, not licenses. A future dependency under a copyleft license could silently enter the runtime tree without any gate flagging the NOTICE drift. Manual-only governance is the gap between A- and A.

---

mvillmow:
## Summary

NOTICE makes correct manual license-compatibility determinations (BSD/MIT/Apache base; LGPL/PSF as non-vendored optional extras), but no CI job verified that a newly added or transitively-upgraded **distributed** dependency stays license-compatible. `pip-audit` (security.yml) checks CVEs, not licenses — so a copyleft dependency could silently enter the shipped tree and drift from NOTICE.

This adds a **stdlib-only drift-guard** + a CI gate that fails when a distributed dependency's license is not BSD-3-Clause-compatible.

## Changes

- **`scripts/check_license_compatibility.py`** — enumerates the DISTRIBUTED dependency set (base deps + runtime extras read from `Requires-Dist`, **never `dev`**), resolves each license across three metadata fields (`License-Expression` → trove `Classifier` → freeform `License`), and checks a two-tier allowlist:
  - blanket `PERMISSIVE` (MIT/BSD/Apache/ISC/PSF-via-Python-2.0…)
  - per-package `ALLOWED_EXTRA_COPYLEFT` (`pygithub`→LGPL, `defusedxml`→PSF) mirroring NOTICE.
  - **Fails loudly** (exit 2) when the package is uninstalled, `Requires-Dist` is empty, or an *installable* declared dep has no metadata — never a silent no-op.
  - **Blocking on `pull_request`** (exit 1), **advisory on main/schedule** (exit 0) via `GITHUB_EVENT_NAME`.
  - Deps absent only because their marker excludes the current interpreter (`tomli` on Py≥3.11, `tzdata` off Windows) are skipped with a note, not treated as coverage holes.
- **`.github/workflows/security.yml`** — `license-scan` job pins the repo-standard `actions/setup-python@a309ff8b…#v6` with `cache: pip`, installs `pip install -e ".[all]"` (every runtime extra), and runs the gate.
- **`NOTICE`** — cross-references the machine-enforced gate and the `ALLOWED_EXTRA_COPYLEFT` map.
- **`tests/unit/scripts/test_check_license_compatibility.py`** — synthetic + real-installed-metadata resolver tests, OR-semantics, two-tier allowlist, loud-failure (uninstalled package / installable-but-absent dep), distributed-scope (dev tools excluded, `tzdata` in scope), and a `[all]`-completeness guard asserting every `RUNTIME_EXTRAS` member is aggregated in pyproject's `[all]` so coverage cannot silently drop.

## Design notes

- **No pre-commit hook.** The check requires the package installed with `.[all]`; a bare dev env lacks it, so a local hook would either crash, block every commit (POLA violation), or silently pass (forbidden). The CI `license-scan` job installs `.[all]` and runs it loudly — it is the single authoritative gate.
- **Bare setup-python + pip, not pixi** — mirrors the unit-test jobs; pixi+pip co-managing site-packages causes lockfile churn.

## Verification

- New suite: 23 passed, 3 env-dependent skips locally (all run in CI's installed env).
- Bare-venv CI scenario (`pip install -e ".[all]"` + run): all installed distributed deps classified compatible; `tomli`/`tzdata` correctly skipped.
- Acceptance: copyleft on PR → exit 1; same on main → exit 0; `MIT OR GPL-3.0` → allowed; PSF allowed for `defusedxml` but rejected elsewhere.
- `pre-commit run` on all changed files: ruff/format/mypy/yamllint/bandit/test-structure all pass.

Closes #1219


scripts/check_license_compatibility.py
```diff
@@ -0,0 +1,313 @@
...
+import importlib.metadata as md
+import json
+import os
+import sys
+
+from packaging.requirements import InvalidRequirement, Requirement
+
+DIST_NAME = "HomericIntelligence-Hephaestus"
+
+# Extras whose packages are DISTRIBUTED. `dev` is deliberately absent.
```
mvillmow, line 39: Allowlist omits the modern SPDX id `LGPL-3.0-only`, which is exactly what NOTICE:37 documents for PyGithub. Today PyGithub resolves via its trove classifier → `LGPL-3.0` (allowlisted), but if a future PyGithub release adds a PEP 639 `License-Expression: LGPL-3.0-only`, `resolve_license` returns `["LGPL-3.0-only"]` (License-Expression wins over trove) → `is_compatible` returns False → the license-scan gate fails on the very dependency it is designed to allow. Add the SPDX-canonical forms: `frozenset({"LGPL-3.0", "LGPL-3.0-only", "LGPL-3.0-or-later", "LGPL"})`. Keeping the allowlist aligned with the spelling NOTICE already uses prevents a routine dependency upgrade from red-flagging CI.

scripts/check_license_compatibility.py
```diff
@@ -0,0 +1,313 @@
...
+RUNTIME_EXTRAS: frozenset[str] = frozenset(
```
mvillmow, line 40: PSF is modeled inconsistently: `Python-2.0` is in the blanket `PERMISSIVE` set (line 30) *and* listed per-package here for `defusedxml`. Two consequences: (1) the per-package `Python-2.0` entry is dead code — any package resolving to `Python-2.0` is already universally allowed by PERMISSIVE; (2) the intended per-package PSF scoping (`test_psf_allowed_only_for_defusedxml` asserts PSF is rejected for other packages) is silently defeated for the `Python-2.0` spelling — only `PSF-2.0` is actually scoped. Pick one canonical PSF id and keep it in a single tier: drop `Python-2.0` from PERMISSIVE if PSF must stay per-package, or drop the per-package `Python-2.0` entry and document that PSF is treated as blanket-permissive.

mvillmow (commented):
Correct, complete, fail-closed license gate; all disk claims verified. Two minor SPDX-id robustness notes (LGPL-3.0-only coverage gap; split PSF modeling).

tests/unit/scripts/test_check_license_compatibility.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+class TestMain:
+    """main() exit-code contract: blocking on PR, advisory on main."""
+
+    def _fixture(self, tmp_path, records):
+        p = tmp_path / "meta.json"
+        p.write_text(json.dumps(records))
+        return str(p)
+
```
mvillmow, line 220: This completeness guard hand-rolls a TOML parse via regex (`^all\s*=\s*\[.*?\[([^\]]+)\]`). It assumes `[all]` stays single-line with the aggregate as the first inner bracket group. If `[all]` is ever reformatted to a multi-line array, gains a comment, or the dist-name placement shifts, the regex silently mis-captures (or `assert m` trips, masking the real coupling it means to enforce). Prefer `tomllib.loads(text)` (stdlib on the project's 3.11+ interpreter) and read `project.optional-dependencies.all` directly — it's both simpler and immune to formatting drift. The prior team learnings (`dependency-manifest-single-source-of-truth`) flag exactly this regex-over-parser pattern.

mvillmow (commented):
GO: license-scan gate correct, fail-closed, well-tested; PR-diff 'deletions' are stale-base artifacts (git diff main = 4 additive files only). One minor: regex TOML parse in a test guard.

mvillmow (commented):

tests/unit/scripts/test_check_license_compatibility.py
```diff
@@ -0,0 +1,252 @@
...
+                "check_license_compatibility.md.metadata",
+                side_effect=md.PackageNotFoundError("tomli"),
+            ):
+                assert scan(None) == []
+
+
+class TestDistributedScope:
+    """distributed_requirements selects the distributed set, excludes dev."""
+
+    def _dist_or_skip(self):
```
mvillmow, line 158: Re-opening: prior review comment not addressed — _dist_or_skip is unchanged: the except block still ends with pytest.skip(...) and no explicit `return set()` follows it. The requested fix was not applied.

> Explicit returns mixed with implicit (fall through) returns. Add an explicit `return set()` immediately after `pytest.skip(...)` in the except block of TestDistributedScope._dist_or_skip.

hephaestus/ci/precommit.py
```diff
@@ -49,19 +49,10 @@
...
-#: Only tools that are *pixi-backed* (have a version floor in pixi.toml) belong
-#: here, so the hook can flag local-vs-CI drift. ``yamllint`` is the mirrored
-#: external hook whose floor pixi.toml owns (see the single-source-of-truth note
-#: at the top of ``.pre-commit-config.yaml``). ``mirrors-mypy`` maps to the
-#: ``mypy`` floor for repos that mirror mypy as an external hook. Meta hooks
-#: like ``pre-commit/pre-commit-hooks`` are intentionally absent: they have no
-#: pixi package (they run in pre-commit's own venv), so requiring a pixi entry
-#: would produce a false MISSING finding.
 DEFAULT_HOOK_TO_PIXI_MAP: dict[str, str] = {
     "https://github.com/pre-commit/mirrors-mypy": "mypy",
```
mvillmow, line 62: Unrelated to #1219 and a behavioral regression: this hunk removes `yamllint` from DEFAULT_HOOK_TO_PIXI_MAP (dropping its rev-drift coverage) and adds `pre-commit/pre-commit-hooks`. The deleted docstring explicitly documented that meta hooks like `pre-commit/pre-commit-hooks` are intentionally absent because they have no pixi package and would produce a false MISSING finding — this change reintroduces exactly that false positive. This is a revert of #1249's intent, carried in by the stale base. It should vanish after rebasing onto main.

mvillmow (commented):
License-scan feature is correct/complete, but the diff reverts merged unrelated features (#1214 build-dir guard, #1249 precommit-version hook, precommit.py mapping). Stale base — rebase onto main before merge.

mvillmow (commented):
LGTM — license-scan gate correct/complete; diff now scoped to 4 files (contamination reverted); completeness test uses tomllib+PEP508. Zero findings.

Issue #435, opened 2026-05-22
fix: check_version_single_source.py reports a false PASS under dynamic versioning


mvillmow:
## Problem

`scripts/check_version_single_source.py:37` uses the regex
`\[project\].*?version\s*=\s*"([^"]+)"` with `re.DOTALL` to find the project version.

But `pyproject.toml` uses `hatch-vcs` dynamic versioning: it declares
`dynamic = ["version"]` and has **no literal `version = "..."` field** under
`[project]`. The DOTALL regex skips past `dynamic = ["version"]` and matches the first
`= "..."` it finds — which is an entry-point string in `[project.scripts]`.

Verified live:

```
OK: pyproject.toml [project] version = "hephaestus.validation.python_version:main"
Version single source of truth: PASS
exit: 0
```

The pre-commit hook `check-version-single-source` has been silently passing on a
nonsense captured value on every run. It validates nothing.

## Fix

The real invariant under hatch-vcs is: `pyproject.toml` must declare `dynamic`
containing `"version"` with `[tool.hatch.version]` `source = "vcs"`, and must NOT have
a static `[project].version`. `pixi.toml` must have no version field.

Rewrite the checker to parse `pyproject.toml` with `tomllib`/`tomli` and validate the
actual dynamic-versioning contract, and add unit tests.

## Acceptance

- The hook FAILS if a static `[project].version` is reintroduced.
- The hook FAILS if `dynamic` no longer contains `"version"` or hatch-vcs source is removed.
- The hook PASSES on the current valid config.
- Unit tests cover all three cases.

Found by the strict 1.0-readiness audit (Section 12, CRITICAL).

PR #1266, opened 2026-06-13
ci(version-sync): make deps/version-sync a real gate

### LINKED ISSUE #1181: [audit][MAJOR] deps/version-sync is a required check that asserts almost nothing
**Location:** `.github/workflows/_required.yml:790-820`

deps/version-sync (.github/workflows/_required.yml:770-819) is a required check that asserts nothing: DYNAMIC is computed and echoed but never used in a conditional, so a 'static' result does not fail. The only exit-1 path is a VERSION-file mismatch, but the exit 1 is actually at line 814 (the `if` comparison is at :812), and no VERSION file exists in the repo, making that branch dead. The job therefore always passes regardless of version configuration. (Only nit: finding cites the exit at :812; it is at :814 — the comparison is at :812.)

---
_Filed by strict full-coverage repo audit (2026-06-12). Verified on disk by an independent skeptic agent before filing. Overall audit: 85.1% CONDITIONAL GO._

mvillmow:
## Summary

The required check `deps/version-sync` (`.github/workflows/_required.yml`) asserted **nothing**: `DYNAMIC` was computed and echoed but never branched on, and its only `exit 1` lived inside a dead `if [ -f VERSION ]` block (no `VERSION` file exists in the repo). The job therefore passed for any version configuration (#1181).

This makes the job genuinely enforce the three-file contract its name and `docs/DEFINITION_OF_DONE.md:28` promise — "pyproject.toml then pixi.toml then pixi.lock".

## Changes

- **Drop `--no-deps`** on the editable install. The single-source checker imports `hephaestus.utils.helpers` → `packaging` (a declared core dep, `pyproject.toml:29`); `--no-deps` would `ModuleNotFoundError` instead of asserting.
- **Legs 1+2** — invoke the already-tested checker `hephaestus.scripts_lib.check_version_single_source` via `python -m` so `get_repo_root()` anchors to the installed package (immune to `sys.path[0]`/CWD `__file__` resolution). Under `set -e`, a violation (static `[project].version`, missing `dynamic`, wrong hatch source, pixi `[workspace].version`) fails the job — the live exit-1 path the dead VERSION branch never provided. Re-uses the checker covered by `tests/unit/scripts_lib/test_check_version_single_source.py` rather than re-deriving in bash (avoids the regex false-PASS class of #435).
- **Leg 3** — add `pixi install --locked` to assert `pixi.lock` is in sync with `pixi.toml`: the DoD-named pixi.lock leg the old job dropped. Uses the repo-pinned pixi `v0.69.0`, the same frozen verification the `pixi-check` job already runs (guaranteed to support `--locked`).
- **Delete** the computed-and-discarded `DYNAMIC` block and the dead `VERSION`-file branch (YAGNI).

## Why not delete the job?

`deps/version-sync` is a **pinned required status check** in the active org ruleset. Deleting the job without de-listing the context would leave every PR permanently `BLOCKED` on a check that never reports. The correct move is to make the pinned gate assert its named contract directly.

## Verification (clean venv mirroring CI: `pip install -e .` + bare `python`)

- Leg 1+2: checker exits **0** on clean `main`, **1** on a synthetic static-`[project].version` violation, **0** restored; `packaging` imports under the exact CI install.
- Leg 3: `pixi install --locked` exits **0** on a clean tree, **1** on a `pixi.toml` drift, **0** restored.
- `yamllint` clean; `check-jsonschema --builtin-schema vendor.github-workflows` → `ok`.
- 21 existing checker unit tests pass.

Closes #1181

### DIFF
--- a/.github/workflows/_required.yml
+++ b/.github/workflows/_required.yml
@@ -820,36 +820,48 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install package (for __version__)
-        run: pip install -e "." --no-deps
+      - name: Install package (for __version__ and checker imports)
+        # Install WITH deps: the single-source checker imports
+        # ``hephaestus.utils.helpers``, which imports ``packaging`` (a declared
+        # core dependency). ``--no-deps`` would skip ``packaging`` and the
+        # checker below would ModuleNotFoundError instead of asserting.
+        run: pip install -e "."
 
-      - name: Check version consistency
+      - name: Check version single-source-of-truth (pyproject + pixi.toml)
         run: |
           set -euo pipefail
 
-          # Extract version from pyproject.toml via hatch-vcs (dynamic)
-          DYNAMIC=$(python -c "
-          import tomllib
-          with open('pyproject.toml', 'rb') as f:
-              d = tomllib.load(f)
-          dynamic = d.get('project', {}).get('dynamic', [])
-          print('dynamic' if 'version' in dynamic else 'static')
-          ")
-          echo "Version strategy: $DYNAMIC"
-
-          # Resolve the actual installed version
+          # Diagnostics (informational only).
           PKG_VERSION=$(python -c "import hephaestus; print(hephaestus.__version__)")
           echo "Installed __version__: $PKG_VERSION"
 
-          # If a VERSION file exists, cross-check it
-          if [ -f VERSION ]; then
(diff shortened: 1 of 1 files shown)


mvillmow (commented):
Dead deps/version-sync gate now enforces all 3 legs (checker via -m + pixi install --locked). Import fix correct, live exit-1 path verified, leg-3 matches proven pixi-check job. GO.

PR #1275, opened 2026-06-13
fix(scripts): isolate gh stderr from jq input in choose_merge_flag

### LINKED ISSUE #1125: Follow-up from #911: 2 item(s) (core, safety)
_Consolidated follow-up from implementation of #911._

### choose_merge_flag.sh:21 mixes gh stderr into jq input

scripts/choose_merge_flag.sh:21 captures `gh api repos/${repo} 2>&1` into $raw, then pipes $raw into `jq -r` at line 26. When `gh` writes any stderr noise (deprecation warning, update banner, rate-limit notice, GH_DEBUG=1 traces) alongside a valid JSON body on stdout, the merged blob is no longer valid JSON; `jq -r` returns empty, and the function falls through to `target repo ${repo} allows no merge methods` (exit 1) even though all three flags would have been valid. Repro: `gh() { echo 'warning: x' >&2; echo '{"allow_rebase_merge":true,"allow_squash_merge":true,"allow_merge_commit":true}'; }; export -f gh; . scripts/choose_merge_flag.sh; choose_merge_flag o/r` yields empty stdout + exit 1 instead of `--rebase`. Verified this exact repro returns `result: []  exit=1`. Impact: on ProjectHephaestus the skill-level `--squash` fallback masks it; on a rebase-only or merge-only target repo, callers either error or arm the wrong merge method. Fix: capture stdout only — `if ! raw=$(gh api "repos/${repo}" 2>/tmp/gh.err); then ... ${raw}=$(cat /tmp/gh.err); fi` (or use a temp file / process substitution to keep stderr separate for the error message). Add a regression test using the gh-mock pattern above.

### Integration tests skip choose_merge_flag positive paths

tests/integration/test_choose_merge_flag_sh.py:8 only covers function definition and tests/integration/test_choose_merge_flag_sh.py:19 only covers missing-arg exit 2. The three behaviors documented in scripts/choose_merge_flag.sh:9-13 — preference order rebase→squash→merge selection, `allows no merge methods` exit 1, successful flag emission — are never exercised by automated tests. The PR-911 plan's Verification section sketched the `bash -c 'gh() { case "$1" in api) echo "{...}";; esac; }; export -f gh; . scripts/choose_merge_flag.sh && choose_merge_flag o/r'` mock-gh pattern but it was not checked in. Add three integration tests using that pattern: (a) all-three-allowed yields `--rebase`, (b) only squash allowed yields `--squash`, (c) none allowed yields exit 1 with stderr containing `allows no merge methods`. These would also catch the safety bug above (any of them would fail if a non-empty stderr leaks into $raw). Qualifies as `core` per the test-scope rule — tests for code shipped in this PR.

---

mvillmow:
Isolates `gh api` stderr from the JSON body that `jq` parses in `choose_merge_flag()`.

## Summary

- Replace `2>&1` with `mktemp` temp file + `trap "rm -f '$_err_file'" RETURN` so deprecation warnings, rate-limit notices, and `GH_DEBUG` traces never corrupt the JSON piped to `jq`
- Add regression guard test (`test_shell_helper_handles_gh_stderr_noise`) that fails against the unfixed script and passes after the fix
- Add three positive-path integration tests: rebase preference order, squash fallback, no-methods exit-1

## Test plan

- [x] TDD RED confirmed: `test_shell_helper_handles_gh_stderr_noise` fails against unfixed `2>&1` version
- [x] TDD GREEN confirmed: all 6 tests pass after fix (`pixi run pytest tests/integration/test_choose_merge_flag_sh.py -v --no-cov`)
- [x] Commit is GPG-signed

Closes #1125

### DIFF
--- a/scripts/choose_merge_flag.sh
+++ b/scripts/choose_merge_flag.sh
@@ -17,9 +17,12 @@ choose_merge_flag() {
         echo "choose_merge_flag: missing required argument <owner/repo>" >&2
         return 2
     fi
-    local raw flag
-    if ! raw=$(gh api "repos/${repo}" 2>&1); then
-        echo "choose_merge_flag: gh api repos/${repo} failed: ${raw}" >&2
+    local raw flag _err_file
+    _err_file=$(mktemp)
+    # shellcheck disable=SC2064
+    trap "rm -f '$_err_file'" RETURN
+    if ! raw=$(gh api "repos/${repo}" 2>"$_err_file"); then
+        echo "choose_merge_flag: gh api repos/${repo} failed: $(cat "$_err_file")" >&2
         echo "  (check: gh auth status; token needs 'repo' scope; repo exists)" >&2
         return 1
     fi
--- a/tests/integration/test_choose_merge_flag_sh.py
+++ b/tests/integration/test_choose_merge_flag_sh.py
@@ -25,3 +25,71 @@ def test_shell_helper_errors_on_missing_arg() -> None:
     )
     assert out.returncode == 2
     assert "missing required argument" in out.stderr
+
+
+def _run_with_mock_gh(json_body: str, repo: str = "owner/repo") -> subprocess.CompletedProcess:  # type: ignore[type-arg]
+    """Run choose_merge_flag with a mock gh function returning json_body."""
+    script = f"""
+gh() {{
+    case "$1" in
+        api) printf '%s\\n' '{json_body}' ;;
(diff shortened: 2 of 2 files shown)


mvillmow (commented):
LGTM — stderr/stdout isolation via mktemp+trap RETURN is correct; 4 tests cover rebase/squash/exit-1 plus a RED-verified safety regression guard. Both issue items delivered.

PR #1278, opened 2026-06-13
fix(automation): exclude post-loop records from loops_run in main()

### LINKED ISSUE #1153: Follow-up from #818: 1 item(s) (core)
_Consolidated follow-up from implementation of #818._

### loops_run inflated by post-loop records in main() at loop_runner.py:1501

After #818, `main()` computes `loops_run = max((r.loop_idx for r in results), default=0)` over the FULL result list, which now includes post-loop RepoResults that carry `loop_idx=cfg.loops` (set in `_run_post_loop_stages` at `loop_runner.py:1215`). When early-exit fires on loop 1 of 5 and drive-green is selected, the post-loop record's `loop_idx=5` is then reported as `loops_run=5` in `emit_json_status(..., loops_run=loops_run)` at `loop_runner.py:1510/1520`, even though the iteration body only completed loop 1. `run_loop` itself correctly filters these out internally at `loop_runner.py:1337` (`per_loop_results = [r for r in all_results if not r.post_loop_phases]`), but `main()` does not apply the same filter. The existing test `TestMainLoopsRunReporting.test_main_loops_run_early_exit` does not catch this because it stubs `run_loop` directly with a pre-built result list that has no post-loop records. Fix: in `main()` at `loop_runner.py:1501`, replace with `loops_run = max((r.loop_idx for r in results if not r.post_loop_phases), default=0)`. Add a regression test that feeds `run_loop` a mixed result list (per-loop with `loop_idx=1` + post-loop with `loop_idx=5`) and asserts `loops_run == 1`.

---

mvillmow:
## Summary

- `main()` in `loop_runner.py` computed `loops_run` over the full `run_loop` result list, which includes post-loop `RepoResult` records carrying `loop_idx=cfg.loops`
- When early-exit fires on loop 1 of 5 and `_run_post_loop_stages` runs, `loops_run` was inflated to 5 instead of 1
- Fix: filter with `not r.is_post_loop` (the dedicated boolean flag at line 1395), mirroring the canonical filter already applied at line 1525 inside `run_loop`
- Adds `TestMainLoopsRunReporting.test_main_loops_run_excludes_post_loop_records` as a regression guard

Closes #1153

### DIFF
--- a/hephaestus/automation/loop_runner.py
+++ b/hephaestus/automation/loop_runner.py
@@ -1692,7 +1692,7 @@ def main(argv: list[str] | None = None) -> int:  # noqa: C901
     results = run_loop(cfg, repos)
 
     # Compute actual loops run (may be less than cfg.loops due to early exit)
-    loops_run = max((r.loop_idx for r in results), default=0)
+    loops_run = max((r.loop_idx for r in results if not r.is_post_loop), default=0)
 
     failures = [r for r in results if r.any_failure]
     if failures:
--- a/tests/unit/automation/test_loop_runner_early_exit.py
+++ b/tests/unit/automation/test_loop_runner_early_exit.py
@@ -711,6 +711,34 @@ class TestMainLoopsRunReporting:
         assert captured["failed_repos"] == []
         assert captured["rc"] == 0
 
+    def test_main_loops_run_excludes_post_loop_records(self) -> None:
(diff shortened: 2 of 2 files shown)


mvillmow (commented):
LGTM — fix at loop_runner.py:1695 mirrors canonical :1525 is_post_loop filter; regression test is genuine RED→GREEN, fixture correctly tags is_post_loop=True. No defects, no scope creep.

PR #1279, opened 2026-06-13
fix(automation): tighten planner-learn record-writing robustness

### LINKED ISSUE #1138: planner-learn record-writing: minor robustness + coverage gaps (follow-up to #1135)
Strict review of the merged PR #1135 ("persist planner learn attempts") surfaced a few **minor** robustness/quality issues in `_write_planner_learn_record` (`hephaestus/automation/planner_review_loop.py`). The code is functionally correct (verified: `_call_claude` raises on empty output rather than returning None, so the `write_text(None)` TypeError concern does not apply), but these are worth tightening:

1. **Write order / atomicity (MINOR).** `_write_planner_learn_record` writes the `.log` before the `.json`, both inside one `except OSError`. If the `.log` write succeeds but the `.json` write fails, state is half-written (a `.log` ghost with no metadata record). Write the authoritative `.json` first, or use a write-then-rename for atomicity.

2. **Missing OSError-swallow test (MINOR).** `test_writes_success_record` / `test_writes_failure_record` cover the happy/failure paths but not the `except OSError` guard that keeps record-writing non-fatal. Add a test: patch `Path.write_text` to raise `OSError`, assert `capture_planner_learnings` still returns its output and does not raise.

3. **Redundant `learn_succeeded_at` (NITPICK).** It always equals `learn_attempted_at` (single timestamp captured at record-write time), so it carries no extra signal. Either drop it or capture a real start timestamp before the `_call_claude` call and add a `learn_duration_s`.

4. **Prompt directive placement (NITPICK).** In `build_learn_prompt`, the new "Do NOT return a plan. Do NOT ask for approval." directives are injected mid-prompt after the (long) context/plan text, and sit alongside the context's existing "reply with only 3-5 brief bullets" instruction. Consider moving the imperative directives to the front of the prompt for clarity.

None of these are blockers; #1135 is already merged and working.

mvillmow:
Four minor robustness/quality improvements from the strict review of #1135:

1. **Write order**: `.json` is now written before `.log` in `_write_planner_learn_record`, so the authoritative metadata record lands first. A partial-write failure can no longer leave a `.log` ghost with no corresponding `.json`.

2. **OSError-swallow test**: Added `test_oswrite_failure_is_nonfatal` — patches `Path.write_text` to raise `OSError` and asserts `capture_planner_learnings` still returns its output without propagating.

3. **`learn_succeeded_at` → `learn_duration_s`**: Replaced the redundant `learn_succeeded_at` field (always equalled `learn_attempted_at`) with a real `learn_duration_s` captured via `time.monotonic()` before the `_call_claude` call.

4. **Prompt directive placement**: Moved "Do NOT return a plan. Do NOT ask for approval." directives to the front of `build_learn_prompt`, before the context detail suffix, so they appear before the long plan text rather than buried after it.

Closes #1138

### DIFF
--- a/hephaestus/automation/learn.py
+++ b/hephaestus/automation/learn.py
@@ -88,15 +88,15 @@ def _write_learn_record(
 def build_learn_prompt(context: str) -> str:
     """Return the standard automation prompt for the user-facing /learn skill."""
     detail = context.strip()
-    if detail:
-        detail = f" {detail}"
+    suffix = f" {detail}" if detail else ""
     return (
-        f"/learn{detail}"
+        "/learn"
         " EXECUTE the /learn skill-creation workflow for ProjectMnemosyne."
         " Do NOT return a plan. Do NOT ask for approval."
         " Commit the results and create a PR."
         " IMPORTANT: Only push skills to ProjectMnemosyne."
         " Do NOT create files under .claude-plugin/ in this repo."
+        f"{suffix}"
     )
 
 
--- a/hephaestus/automation/planner_review_loop.py
+++ b/hephaestus/automation/planner_review_loop.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import json
 import logging
+import time
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Protocol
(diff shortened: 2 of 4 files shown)


mvillmow (commented):
LGTM — all 4 items (write order, OSError test, learn_duration_s, directive placement) correctly implemented with proportional tests; no regressions to empty-context callers.

PR #1303, opened 2026-06-13
fix(license-scan): classify marker-excluded deps via FALLBACK_LICENSES

### LINKED ISSUE #1256: Follow-up from #1219: 1 item(s) (core)

The license-scan job (.github/workflows/security.yml) runs only python-version 3.13 on ubuntu-24.04. scripts/check_license_compatibility.py:265-270 skips-with-note any in-scope dep whose marker excludes the current interpreter. Result: `tomli` (pyproject.toml:59,65 — gated python_version<'3.11') and `tzdata` (pyproject.toml:34 — gated platform_system=='Windows') are in the distributed scope set but are NEVER license-classified by any CI row — a permanent silent coverage hole in the gate that #1219 built to eliminate silent holes (both are MIT today, so no current violation, but a future transitive upgrade of either would not be caught). Sketch fix: either (a) add a second license-scan matrix leg (e.g. python-version 3.10 to cover tomli, and document tzdata as Windows-unreachable-on-Linux) so every distributed dep is classified by at least one row, or (b) add a static license fallback map for markered-out deps (tomli->MIT, tzdata->MIT from NOTICE) so resolve_license can classify them without an install. Lock with a test asserting the union of classified deps across CI rows covers the full distributed_requirements set.

---

mvillmow:
## Summary

- `tomli` (`python_version<'3.11'`) and `tzdata` (`platform_system=='Windows'`) were in the distributed dependency scope but silently skipped on the single Python 3.13 / ubuntu-24.04 CI leg — a permanent license-classification blind spot.
- Add `FALLBACK_LICENSES: dict[str, list[str]]` (NOTICE-backed literals) so `scan()` classifies marker-excluded deps without a metadata install.
- `scan()` now fails closed (`sys.exit(2)`) if a marker-excluded dep has no `FALLBACK_LICENSES` entry, ensuring any future gated dep is caught.
- Replace the now-incorrect `test_uninstalled_other_python_dep_skipped_not_failed` with three targeted tests; add `TestFallbackLicenses` coverage-completeness lock.

## Test plan

- [ ] `TestLoudFailure::test_marker_excluded_dep_classified_from_fallback` — tomli MIT classifies as no-violation
- [ ] `TestLoudFailure::test_marker_excluded_dep_not_in_fallback_exits_nonzero` — missing fallback → exit 2
- [ ] `TestLoudFailure::test_marker_excluded_dep_with_incompatible_fallback_is_violation` — incompatible fallback surfaces as violation
- [ ] `TestFallbackLicenses::test_fallback_map_entries_are_compatible` — all map entries are permissive
- [ ] `TestFallbackLicenses::test_tzdata_in_fallback` / `test_tomli_in_fallback_on_py311_plus` — concrete assertions on both deps
- [ ] Full unit suite passes with no regressions

Closes #1256

### DIFF
--- a/scripts/check_license_compatibility.py
+++ b/scripts/check_license_compatibility.py
@@ -52,6 +52,17 @@ ALLOWED_EXTRA_COPYLEFT: dict[str, frozenset[str]] = {
     "defusedxml": frozenset({"PSF-2.0", "Python-2.0"}),
 }
 
+# Static license fallback for distributed deps whose PEP 508 marker excludes
+# the CI interpreter/platform (tomli: python_version < '3.11'; tzdata:
+# platform_system == 'Windows'). These packages are never installed in the
+# current CI leg so their metadata is not resolvable; licenses are taken from
+# NOTICE, which is the human-maintained authoritative record. Any future
+# incompatible license change MUST update both NOTICE and this map.
+FALLBACK_LICENSES: dict[str, list[str]] = {
+    "tomli": ["MIT"],  # NOTICE: "toml extra / tomli  MIT"
+    "tzdata": ["Apache-2.0"],  # NOTICE: "tzdata  Apache-2.0"
+}
+
 TROVE_TO_SPDX: dict[str, str] = {
     "MIT License": "MIT",
     "BSD License": "BSD-3-Clause",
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
LGTM — FALLBACK_LICENSES closes the tomli/tzdata silent coverage hole; scan() fails-closed on uncovered gated deps; tests lock compatibility+completeness. NOTICE values verified.

PR #1304, opened 2026-06-13
fix(license-scan): classify marker-excluded deps via static fallback

### LINKED ISSUE #1258: Follow-up from #1219: 1 item(s) (core)

The license-scan job (.github/workflows/security.yml) runs only python-version 3.13 on ubuntu-24.04. scripts/check_license_compatibility.py:scan() skips-with-note any in-scope dep whose marker excludes the current interpreter (the `installable_now=False` branch). Result: `tomli` (pyproject.toml gated python_version<'3.11') and `tzdata` (pyproject.toml gated platform_system=='Windows') are in the distributed scope set but are NEVER license-classified by any CI row — a permanent silent coverage hole in the gate that #1219 built to eliminate silent holes. Both are MIT today so no current violation, but a future transitive upgrade of either would not be caught. Sketch fix: either (a) add a second license-scan matrix leg (python-version 3.10 to cover tomli; document tzdata as Windows-unreachable-on-Linux) so every distributed dep is classified by at least one row, or (b) add a static fallback license map for markered-out deps (tomli->MIT, tzdata->MIT per NOTICE) so resolve_license can classify them without an install. Lock with a test asserting the union of classified deps across CI rows covers the full distributed_requirements set.

---

mvillmow:
## Summary

- `tomli` (python_version<3.11) and `tzdata` (platform_system==Windows) are distributed deps that are never installed on the single CI leg (Python 3.13/Linux), creating a permanent silent coverage hole in the license gate built by #1219.
- Add `STATIC_FALLBACK_LICENSES` dict with SPDX IDs sourced from NOTICE (`tomli`→MIT, `tzdata`→Apache-2.0).
- The silent-skip branch now looks up the fallback map; deps absent from the map `exit(2)` — a future marker-excluded dep cannot slip through without an explicit NOTICE-backed entry.
- Update the module docstring `FAILS LOUDLY` block and rename the corresponding test.
- Add `TestStaticFallback` (6 tests) including two staleness mitigations: value cross-check against real `importlib.metadata` (runs when dep is installed) and key+SPDX cross-check against NOTICE (runs unconditionally on every platform).

## Test plan

- [x] `pixi run pytest tests/unit/scripts/test_check_license_compatibility.py -v --no-cov` — 29 passed, 5 skipped
- [x] `pixi run pytest tests/unit/scripts/ -v --no-cov` — 125 passed, 5 skipped
- [x] Signed commit verified via `git log --show-signature -1`

Closes #1258

### DIFF
--- a/tests/unit/scripts/test_check_license_compatibility.py
+++ b/tests/unit/scripts/test_check_license_compatibility.py
@@ -350,8 +350,10 @@ class TestStaticFallback:
                 f"{pkg!r} is in STATIC_FALLBACK_LICENSES but not mentioned in NOTICE — "
                 "NOTICE is the authoritative source; add the dep there first."
             )
+            pkg_lines = [line for line in notice_text.splitlines() if pkg.lower() in line.lower()]
             for spdx_id in spdx_ids:
-                assert spdx_id in notice_text, (
+                assert any(spdx_id in line for line in pkg_lines), (
                     f"STATIC_FALLBACK_LICENSES[{pkg!r}] = {spdx_ids!r} but SPDX id "
-                    f"{spdx_id!r} not found in NOTICE — update one to match the other."
+                    f"{spdx_id!r} not found on any NOTICE line mentioning {pkg!r} — "
+                    "update one to match the other."
                 )
--- a/tests/unit/scripts/test_check_license_compatibility.py
+++ b/tests/unit/scripts/test_check_license_compatibility.py
@@ -17,6 +17,7 @@ from check_license_compatibility import (
     ALLOWED_EXTRA_COPYLEFT,
     DIST_NAME,
     RUNTIME_EXTRAS,
+    STATIC_FALLBACK_LICENSES,
     _FixtureMeta,
     distributed_requirements,
     is_compatible,
@@ -139,9 +140,9 @@ class TestLoudFailure:
                     scan(None)
         assert exc.value.code == 2
 
-    def test_uninstalled_other_python_dep_skipped_not_failed(self):
-        # installable_now=False (marker excludes this interpreter, e.g. tomli on
-        # Python >= 3.11) => correctly absent => skipped, not a coverage hole.
+    def test_uninstalled_other_python_dep_with_fallback_classifies_not_fails(self):
+        # installable_now=False + known static fallback => classified compatible,
+        # not a coverage hole. tomli's fallback is MIT.
         with patch(
             "check_license_compatibility.distributed_requirements",
             return_value=[("tomli", False)],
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
GO: NOTICE-sourced static fallback closes the tomli/tzdata gate hole + coverage-lock test; one minor: NOTICE value cross-check not bound to the package line.

tests/unit/scripts/test_check_license_compatibility.py
```diff
@@ -257,3 +258,100 @@ def test_pygithub_lgpl_fixture_passes(self, tmp_path):
...
+        for pkg, spdx_ids in STATIC_FALLBACK_LICENSES.items():
+            assert pkg in notice_text.lower(), (
+                f"{pkg!r} is in STATIC_FALLBACK_LICENSES but not mentioned in NOTICE — "
+                "NOTICE is the authoritative source; add the dep there first."
+            )
+            for spdx_id in spdx_ids:
+                assert spdx_id in notice_text, (
+                    f"STATIC_FALLBACK_LICENSES[{pkg!r}] = {spdx_ids!r} but SPDX id "
+                    f"{spdx_id!r} not found in NOTICE — update one to match the other."
+                )
```
mvillmow, line 359: The SPDX value check (`assert spdx_id in notice_text`) is decoupled from the package: `Apache-2.0` also appears on NOTICE's `packaging` line, so tzdata's value could drift to a wrong-but-present SPDX id and this test would still pass. The per-key check above (`pkg in notice_text.lower()`) is sound; only the value binding is loose. The stronger per-package check (`test_static_values_match_installed_metadata`) skips on Py3.13/Linux, so on the actual CI leg neither value is verified against an authoritative per-package source. Suggest scoping the value assertion to the package's NOTICE line, e.g. find the line containing `pkg` (case-insensitive) and assert the SPDX id appears on that line rather than anywhere in the file.

mvillmow (commented):
LGTM — NOTICE-sourced static fallback closes the tomli/tzdata gate hole, coverage-lock + per-package-line NOTICE cross-check; prior minor resolved.

mvillmow:
## Rebase analysis: superseded

Rebased `1258-auto-impl` onto `origin/main` and found that **all commits silently dropped** — the rebased HEAD is identical to `origin/main` (zero diff).

Main already contains this PR's feature:
- `scripts/check_license_compatibility.py:64` — `FALLBACK_LICENSES` dict with `tomli` (MIT) and `tzdata` (Apache-2.0) already present
- `scripts/check_license_compatibility.py:281` — `FALLBACK_LICENSES.get(pkg)` fallback path already in the scan loop
- `tests/unit/scripts/test_check_license_compatibility.py:252` — `TestFallbackLicenses` class with coverage tests already in main

The marker-excluded dep classification via static fallback (issue #1258) was incorporated into main via a sibling PR. Main uses the variable name `FALLBACK_LICENSES`; this branch used `STATIC_FALLBACK_LICENSES` — the functionality is semantically identical.

**Recommendation: close without merging** — the feature is already live on `main`.

mvillmow:
Superseded: rebasing onto main drops both commits as 'already upstream' — main already contains FALLBACK_LICENSES static fallback. Closing.

PR #1308, opened 2026-06-13
fix(scripts_lib): support YAML sequence format in extract_ci_matrix_python_versions

### LINKED ISSUE #1284: Follow-up from #1189: 1 item(s) (core)
_Consolidated follow-up from implementation of #1189._

### extract_ci_matrix_python_versions misses YAML sequence format

hephaestus/validation/python_version.py:290-293 — _CI_MATRIX_PYTHON_RE only matches inline bracket lists (`python-version: ["3.10", "3.11"]`). GitHub Actions also supports multi-line YAML sequence format:

```yaml
python-version:
  - "3.10"
  - "3.11"
```

Sketch fix: after the bracket-list match fails, fall back to `re.findall(r'python-version:\s*\n(?:  - "?(\d+\.\d+)"?\n)+', content)` or parse the YAML with `yaml.safe_load`. This is public API (exported in `validation/__init__.py:105`) so callers may pass real workflow content from repos using sequence format.

---

mvillmow:
Fixes `extract_ci_matrix_python_versions` in `hephaestus/scripts_lib/check_python_version_consistency.py` to handle the multi-line YAML sequence format used by many GitHub Actions workflows:

```yaml
python-version:
  - "3.10"
  - "3.11"
```

Previously, only the inline bracket format was matched. When a workflow used the sequence format, `extract_ci_matrix_python_versions` silently returned `[]`, causing `check_ci_matrix_coverage` to short-circuit at its early-return guard (`:200-202`) and report OK despite having no data to compare against.

## Changes

- **`hephaestus/scripts_lib/check_python_version_consistency.py`**: Adds three compiled module-level constants (`_CI_MATRIX_BRACKET_RE`, `_CI_MATRIX_SEQUENCE_RE`, `_CI_VERSION_RE`) and rewrites `extract_ci_matrix_python_versions` with bracket-first / sequence-fallback logic. Both paths share `_CI_VERSION_RE` (DRY). `(?=\n|$)` lookahead in the sequence regex correctly handles the final list item whether or not a trailing newline follows.
- **`tests/unit/scripts_lib/test_check_python_version_consistency.py`**: Adds 7 new unit tests to `TestExtractCiMatrixPythonVersions` (double-quoted, unquoted, single-quoted, dedup+sort, bracket-precedence, 4-space-indent, no-trailing-newline) and 1 integration test to `TestCheckCiMatrixCoverage` verifying the sequence format propagates through the full coverage check.

Closes #1284

### DIFF
--- a/hephaestus/scripts_lib/check_python_version_consistency.py
+++ b/hephaestus/scripts_lib/check_python_version_consistency.py
@@ -14,6 +14,12 @@ from pathlib import Path
 
 from hephaestus.utils.helpers import get_repo_root as _get_repo_root
 
+_CI_MATRIX_BRACKET_RE = re.compile(r"python-version:\s*\[([^\]]+)\]")
+_CI_MATRIX_SEQUENCE_RE = re.compile(
+    r"python-version:\s*\n((?:[ \t]+-\s*[\"']?\d+\.\d+[\"']?(?=\n|$)\n?)+)"
+)
+_CI_VERSION_RE = re.compile(r'["\']?(\d+\.\d+)["\']?')
+
 
 def get_repo_root() -> Path:
     """Find repository root, anchored to this module's location.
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
LGTM — correct sequence-format regex fallback with verified no-trailing-newline handling; 8 TDD tests cover all formats; fixes the real silent-skip in check_ci_matrix_coverage; no scope creep.

PR #1311, opened 2026-06-13
fix(github): guard GITHUB_REPOSITORY env read in severity_label.main()

### LINKED ISSUE #1300: Follow-up from #1210: 1 item(s) (core)
_Consolidated follow-up from implementation of #1210._

### hephaestus-label-severity --version crashes with KeyError when GITHUB_REPOSITORY unset

severity_label.py:145 does `os.environ["GITHUB_REPOSITORY"]` unconditionally. Running `hephaestus-label-severity --version` (or any invocation outside the workflow) without the env var raises `KeyError: 'GITHUB_REPOSITORY'` instead of printing the version and exiting. All other env reads use `.get()` with a fallback or are guarded by the `isdigit()` check. Fix: move the `repo = os.environ[...]` read below the `args = parser.parse_args(argv)` call that already handles `--help`/`--version` exits, or use `os.environ.get("GITHUB_REPOSITORY", "")` and validate it before use (mirroring the `ISSUE_NUMBER` pattern at line 146).

---

mvillmow:
Replace the bare `os.environ["GITHUB_REPOSITORY"]` at `severity_label.py:141`
with `os.environ.get("GITHUB_REPOSITORY", "")` plus an explicit validation
guard, so that invoking `hephaestus-label-severity` outside the workflow exits
cleanly with an error message instead of raising `KeyError`.  The guard mirrors
the existing `ISSUE_NUMBER` pattern immediately below it.

Two new unit tests cover the missing-var and malformed-var paths:
- `test_main_rejects_missing_github_repository`
- `test_main_rejects_malformed_github_repository`

Also fixes merge artefacts introduced when rebasing onto `1210-auto-impl`:
duplicate `## Console-Script Stability Tiers` section in `COMPATIBILITY.md`,
duplicate table rows, stale script count (47→48), and README/docs count drift.

Closes #1300

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -42,3 +42,37 @@ body:
         - hephaestus: 0.3.0
     validations:
       required: false
+  - type: dropdown
+    id: severity
+    attributes:
+      label: Severity
+      description: >
+        How impactful is this? Selecting a value auto-applies the matching
+        `severity:*` label via the auto-label-severity workflow, feeding the
+        Epic-and-children planning pipeline (see docs/ROADMAP.md). Audit-section
+        tagging is handled by maintainers, not this form.
+      options:
+        - critical
+        - major
+        - minor
+        - nitpick
+    validations:
+      required: false
+  - type: input
+    id: parent_epic
+    attributes:
+      label: Parent Epic (optional)
+      description: >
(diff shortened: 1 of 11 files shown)


mvillmow (commented):
LGTM — #1300 delta (GITHUB_REPOSITORY guard + 2 tests) is correct, complete, idiomatic, and verified on disk; KeyError path eliminated.

PR #1312, opened 2026-06-13
docs(license-scan): document marker-excluded fallback path in module docstring

### LINKED ISSUE #1306: Follow-up from #1258: 1 item(s) (core)
_Consolidated follow-up from implementation of #1258._

### scan() docstring falsely says marker-excluded deps are skipped

scripts/check_license_compatibility.py:248-251: the docstring still reads 'Deps absent only because their marker excludes the current interpreter … are skipped with a note, not treated as a hole.' After PR #1304 this is false — they are now classified via STATIC_FALLBACK_LICENSES or exit(2). Stale docstring misleads future contributors adding new marker-excluded deps about the actual behaviour. Fix: update the docstring to describe the fallback-or-exit(2) path.

---

mvillmow:
The module-level `FAILS LOUDLY` block in `scripts/check_license_compatibility.py` omitted the marker-excluded dep case introduced by PR #1303. Contributors reading the top of the file had no indication that marker-excluded deps are classified via `FALLBACK_LICENSES` or exit 2 — the behavior only appeared in the `scan()` function docstring. This adds a fourth bullet to the module summary mirroring the existing `scan()` language.

No code change. Docstring-only.

Closes #1306

### DIFF
--- a/scripts/check_license_compatibility.py
+++ b/scripts/check_license_compatibility.py
@@ -15,6 +15,9 @@ FAILS LOUDLY (never silently passes):
     -> coverage hole -> exit 2. CI installs `.[all]` on a setup-python runner first.
   * a license string not in the mapping tables -> exit (PR:1 / main:0) with a
     hint to extend TROVE_TO_SPDX / LICENSE_ALIASES.
+  * a distributed dep whose PEP 508 marker excludes the current interpreter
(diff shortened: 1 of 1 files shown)


mvillmow (commented):
LGTM — module FAILS LOUDLY block now accurately documents marker-excluded FALLBACK_LICENSES-or-exit(2) behavior, matching scan() contract. Docs-only, scope-confined.

Issue #1267, opened 2026-06-13
Follow-up from #1181: 1 item(s) (core)


mvillmow:
_Consolidated follow-up from implementation of #1181._

Each section below lists scope-checked follow-up items discovered during implementation. Items are restricted to core library defects, security, safety hazards, or critical functional bugs.

### Test get_repo_root first-match-up behavior with nested pyproject markers

During #1181 (PR #1266) the version checker silently validated the WRONG pyproject.toml: hephaestus/utils/helpers.py:99-127 get_repo_root() walks up returning the FIRST ancestor containing `.git` OR `pyproject.toml`, so in a git worktree nested under a parent repo that is itself a Python project, it resolved to the parent meta-repo's pyproject.toml — the checker passed despite an injected static `[project].version` violation. The CI step was worked around via `python -m` (PR #1266), but the resolver's first-match-up contract has NO test for the nested-marker case: tests/unit/utils/test_general_utils.py:133-156 only covers finds_git_repo / no-git-returns-start / string-path / cwd-default. Sketch fix: add a test that builds tmp_path/outer/pyproject.toml and tmp_path/outer/inner/.git (or inner/pyproject.toml), calls get_repo_root(seed=inner_subdir), and asserts it stops at the INNERMOST marker (the worktree root), pinning the documented first-match-up semantics so future regressions in this core resolver are caught.

---

_The implementer also considered the items below and rejected them as out of scope; they are recorded in the PR body, not filed as separate issues._

mvillmow:
# Implementation Plan


The skill file already landed in ProjectMnemosyne. Now producing the revised plan addressing the NOGO finding.

# Implementation Plan

## Objective

Add four focused test cases to `tests/unit/utils/test_general_utils.py` that pin the first-match-up (innermost-wins) contract of `get_repo_root()` (helpers.py:99-127) for nested-marker scenarios. The resolver calls `Path(start_path).resolve()` at helpers.py:117, so all returned paths are resolved; assertions must compare against `.resolve()`d expected paths to be CI-stable on symlinked `$TMPDIR` environments. No production code changes are needed.

## Approach

[...]

PR #1316, opened 2026-06-13
fix(utils): resolve Path.cwd() fallback and symlink-fragile assertions in get_repo_root

### LINKED ISSUE #1309: Follow-up from #1267: 2 item(s) (core)
_Consolidated follow-up from implementation of #1267._

### Fix unresolved fallback path in get_repo_root when no marker found

helpers.py:117 resolves start_path via Path(start_path).resolve() when it is not None, but when start_path is None it assigns Path.cwd() without resolving. The fallback return at helpers.py:127 returns start_path, which is unresolved when start_path was None — producing an inconsistent return type (resolved on success, potentially unresolved on fallback). Fix: change line 117 to `start_path = Path.cwd().resolve() if start_path is None else Path(start_path).resolve()` so the fallback always returns a resolved path.

### Fix symlink-fragile assertions in pre-existing TestGetRepoRoot tests

tests/unit/utils/test_general_utils.py:140 (test_finds_git_repo) asserts `result == mock_git_repo` and line 152 (test_accepts_string_path) asserts `result == mock_git_repo` — neither calls .resolve() on the expected value. get_repo_root() returns a resolved path (helpers.py:122), so on macOS where $TMPDIR is a symlink these assertions fail. Fix: change both assertions to `== mock_git_repo.resolve()`. This is the same class of fragility that issue #1267 was filed to address.

---

mvillmow:
Fix two defects in `get_repo_root` and its test suite, both follow-ups from #1267.

## Changes

- **`hephaestus/utils/helpers.py:117`** — add `.resolve()` to the `Path.cwd()` branch so both arms of the ternary produce a resolved path. The fallback return at line 127 previously returned an unresolved path when `start_path` was `None`, violating the docstring contract ("returns the resolved start path as a fallback").
- **`tests/unit/utils/test_general_utils.py:140`** — add `.resolve()` to expected value in `test_finds_git_repo` to fix symlink-fragile assertion (macOS `$TMPDIR` is a symlink).
- **`tests/unit/utils/test_general_utils.py:152`** — same fix in `test_accepts_string_path`. Both are now consistent with lines 167/179/188 in the same class.
- **`tests/unit/utils/test_general_utils.py:157`** — add `assert result.is_absolute()` to `test_uses_cwd_when_none` as the regression guard for the resolved-path contract (prescribed by the issue body).

## Out of scope

Items considered and rejected as out of scope are recorded in the issue body, not filed as separate issues.

Closes #1309

### DIFF
--- a/hephaestus/utils/helpers.py
+++ b/hephaestus/utils/helpers.py
@@ -114,7 +114,7 @@ def get_repo_root(start_path: str | Path | None = None) -> Path:
         fallback.
 
     """
-    start_path = Path.cwd() if start_path is None else Path(start_path).resolve()
+    start_path = Path.cwd().resolve() if start_path is None else Path(start_path).resolve()
 
     path = start_path
     while path != path.parent:  # Stop at filesystem root
--- a/tests/unit/utils/test_general_utils.py
+++ b/tests/unit/utils/test_general_utils.py
@@ -137,7 +137,7 @@ class TestGetRepoRoot:
         subdir = mock_git_repo / "src" / "module"
         subdir.mkdir(parents=True)
         result = get_repo_root(subdir)
-        assert result == mock_git_repo
+        assert result == mock_git_repo.resolve()
 
     def test_returns_start_path_when_no_git(self, tmp_path):
         """Returns start_path when no .git found."""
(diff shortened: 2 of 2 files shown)


mvillmow (commented):
LGTM — both #1309 criteria implemented correctly; resolved-fallback fix + symlink-safe assertions + is_absolute regression guard. No findings.

PR #1314, opened 2026-06-13
fix(lint): clear 4 pre-commit violations blocking main

mvillmow:
## Summary

`main` was RED — the `lint` job failed on every push due to four pre-commit violations present on `main`. This PR clears all four so `main` goes green again.

| # | Hook | Fix |
|---|------|-----|
| 1 | ruff-format | reformat `tests/unit/scripts_lib/test_check_python_version_consistency.py` |
| 2 | ruff-check E501 | extract `merge_settings` var in `tests/integration/test_choose_merge_flag_sh.py:60` |
| 3 | ruff-check E501 | split docstring in `tests/unit/automation/test_planner_loop.py:681` |
| 4 | mypy unused-ignore | parameterise `CompletedProcess[str]`, drop stale `# type: ignore[type-arg]` in `test_choose_merge_flag_sh.py:30` |

The emitted shell heredoc text is byte-identical (only the Python-side string assembly changed).

## Root cause

- **A** — PR #1294 raised the ruff floor `0.1.x → 0.15`; violations #2/#4 merged earlier (#1275, #1279) under the old ruleset and now fail retroactively.
- **B** — `lint` is **not** a required status check (only the two `test` contexts are), so violation #1 (#1308) merged with red lint.

This PR fixes the symptom. The structural prevention (aggregator gate making lint blocking + branch-protection-as-code) is a separate PR.

## Verification

```
ruff format --check tests/   → 213 files already formatted
ruff check hephaestus/ scripts/ tests/ → All checks passed!
mypy → Success: no issues found in 389 source files
pre-commit run --all-files → all hooks Passed
pytest (3 affected files) → 105 passed
```

Closes #1313

Issue #1313, opened 2026-06-13
fix(lint): main is red — 4 pre-commit violations block all pushes


mvillmow:
## Problem

`main` is currently RED. The `lint` job in the `Required Checks` workflow fails on every push because four pre-commit violations are present on `main` (reproduced locally on the pinned toolchain, ruff 0.15.12):

| # | Hook | File:line | Issue |
|---|------|-----------|-------|
| 1 | ruff-format | `tests/unit/scripts_lib/test_check_python_version_consistency.py:319-322` | string-concat collapse (0.15 formatter) |
| 2 | ruff-check `E501` | `tests/integration/test_choose_merge_flag_sh.py:60` | 110 > 100 |
| 3 | ruff-check `E501` | `tests/unit/automation/test_planner_loop.py:681` | 106 > 100 (docstring) |
| 4 | mypy `unused-ignore` | `tests/integration/test_choose_merge_flag_sh.py:30` | stale `# type: ignore[type-arg]` |

## Root cause (summary)

Two reinforcing causes:
- **A** — PR #1294 raised the ruff floor `0.1.x → 0.15`; violations #2/#4 merged earlier (#1275, #1279) under the old ruleset and now fail retroactively.
- **B** — the `lint` job is **not** a required status check (only the two `test` contexts are), so violation #1 (#1308) merged with red lint.

This issue covers the **immediate fix** (all 4 violations). The structural prevention (make lint blocking via an aggregator gate + branch-protection-as-code) is tracked separately.

mvillmow:
# Implementation Plan


Good — mypy confirms the `type: ignore[type-arg]` at line 30 is unused. The fix is simply to remove it. Now I have all information needed.

# Implementation Plan

## Objective

Fix four pre-commit violations that have left `main` red since ruff's floor was raised from 0.1.x → 0.15 in PR #1294. The violations span three test files: a ruff-format string-concat collapse (violation #1), two E501 line-length overruns (violations #2, #3), and a stale `# type: ignore[type-arg]` that mypy now rejects as unused (violation #4). No logic changes — pure formatting and annotation cleanup.

## Approach

Each violation is mechanically fixed in-place; no new abstractions needed:

- **Violation #1 (ruff-format, `test_check_python_version_consistency.py:321-324`)**: `pyproject.write_text(` with two adjacent string literals. ruff 0.15 collapses implicit concatenation. Fix: join into a single string with a `\n` separator.
[...]

Issue #1173, opened 2026-06-12
fix(resilience): main pre-commit fails on unformatted test_circuit_breaker.py


mvillmow:
The Pre-commit CI workflow is failing on `main` (since the merge of #959) because `tests/unit/resilience/test_circuit_breaker.py` is not ruff-formatted.

## Root cause

PR #959 introduced unformatted code in that file. It merged because the required status checks for `main` are only the unit+integration `test` contexts — `Pre-commit`/`ruff-format` is NOT a required check. The unformatted file therefore landed on main, where the all-files pre-commit gate now fails and contaminates every subsequent PR (e.g. #1061).

This was NOT an admin push; the merges were legitimate auto-merges that simply were not gated on pre-commit.

## Fix

Run `ruff format` on the one offending file (9-line reformat). Follow-up (separate): consider adding `Pre-commit` to the required status checks so this cannot recur.

PR #1317, opened 2026-06-13
harden(ci): make lint blocking via required-checks-gate aggregator + branch-protection-as-code

### LINKED ISSUE #1315: harden(ci): make lint a required check via aggregator gate + branch-protection-as-code
## Problem (root cause of #1313)

main went red because a `lint` failure could reach `main`. Two reinforcing causes:

- **A** — PR #1294 raised the ruff floor `0.1.x → 0.15`, making already-merged code fail retroactively.
- **B** — the `lint` job (and the rest of the `Required Checks` workflow) is **NOT** a required status check. Branch protection's `required_status_checks` lists only:
  - `test (ubuntu-latest, 3.12, unit)`
  - `test (ubuntu-latest, 3.12, integration)`

  So any PR with red lint can merge as long as the two `test` contexts pass. `strict: false` also lets PRs merge against stale `main`.

**Structural enabler:** branch protection is not configured as code, so the gap was invisible.

## Fix

1. Add a single `required-checks-gate` aggregator job to `_required.yml` that `needs:` every gating job (except advisory `auto-merge-policy`) and fails on any non-success/non-skipped result. Uses the `if: always()` + result-check pattern so it does not deadlock on label/auto-merge events where heavy jobs legitimately skip.
2. Make `required-checks-gate` + the two `test` contexts the required status checks, with `strict: true` (applied via `gh api` by an admin — documented in-repo).
3. Add a guard unit test that fails if a new job is added to `_required.yml` without being wired into the gate.
4. Document the required-check contract in `docs/ci/required-checks.md` and cross-link from `docs/DEFINITION_OF_DONE.md`.

This makes "lint failure reaches main" structurally impossible and turns the silent gap into an enforced, documented invariant.

mvillmow:
## Summary

Closes the **root cause** behind #1313: a `lint` failure could reach `main` because branch protection's `required_status_checks` listed only the two `test (...)` contexts — the `lint` job (and the rest of `_required.yml`) was advisory.

This PR makes every gating job actually block merges, via a single aggregator.

### Changes

- **`_required.yml`** — new `required-checks-gate` job. It `needs:` every gating job and reports one aggregated result. Uses `if: always()` + a stdlib-`json` result check so it does **not** deadlock on label/auto-merge PR events where heavy jobs legitimately `skip` (PASS on `success`/`skipped`, FAIL on `failure`/`cancelled`). Advisory `auto-merge-policy` is intentionally excluded.
- **`tests/unit/ci/test_required_checks_gate.py`** — guard test (5 cases) that fails if a job is added to `_required.yml` without being wired into the gate, if the gate gains a dependency it shouldn't have, or if it loses `if: always()`. **Verified it catches an ungated dummy job** (added one locally → test went red → reverted → green).
- **`docs/ci/required-checks.md`** — the required-context contract, why a single aggregator is used (rename/add-safe), the "adding a new gating job" runbook, and the exact `gh api` command to (re-)apply branch protection.
- **`docs/DEFINITION_OF_DONE.md`** — cross-link to the new doc.

### Out-of-band step (admin, after merge)

Branch protection lives on GitHub, not git. After this lands and the gate has reported once, an admin runs (documented in `docs/ci/required-checks.md`):

```bash
gh api -X PUT repos/HomericIntelligence/ProjectHephaestus/branches/main/protection/required_status_checks \
  -f strict=true \
  -f 'checks[][context]=required-checks-gate' \
  -f 'checks[][context]=test (ubuntu-latest, 3.12, unit)' \
  -f 'checks[][context]=test (ubuntu-latest, 3.12, integration)'
```

`strict=true` also closes the "merge against stale `main`" half of the root cause.

### Stacking note

This PR is **stacked on #1314** (base branch `1313-fix-lint-main-red`) so its `tests/` tree is lint-clean. Merge #1314 first; GitHub will retarget this PR to `main` automatically.

Closes #1315

### DIFF
--- a/docs/DEFINITION_OF_DONE.md
+++ b/docs/DEFINITION_OF_DONE.md
@@ -34,6 +34,12 @@ A piece of work is **done** when every item below is true.
 | 21 | Pre-commit hooks pass on the diff | CI job `pre-commit` |
 | 22 | Every review thread is resolved (including bot-authored threads) | Org ruleset `required_review_thread_resolution` |
 
+> **Which of these actually block the merge button?** All of the CI jobs above
+> are aggregated into a single required status check, `required-checks-gate`.
+> See [`docs/ci/required-checks.md`](ci/required-checks.md) for the exact
+> required-context list, why a single aggregator is used, and how to (re-)apply
+> branch protection.
+
 ## For new features
 
 In addition to the universal checklist:
--- a/docs/DEFINITION_OF_DONE.md
+++ b/docs/DEFINITION_OF_DONE.md
@@ -31,7 +31,7 @@ A piece of work is **done** when every item below is true.
 | 18 | Markdownlint passes on all `.md` changes | CI job `markdownlint` |
 | 19 | Shellcheck passes on all shell scripts | CI job `shellcheck` |
 | 20 | Yamllint passes on all YAML changes | CI job `lint` |
-| 21 | Pre-commit hooks pass on the diff | CI job `pre-commit` |
+| 21 | Pre-commit hooks pass on the diff | CI job `lint` (pre-commit suite folded into `lint` per #1173) |
 | 22 | Every review thread is resolved (including bot-authored threads) | Org ruleset `required_review_thread_resolution` |
 
 > **Which of these actually block the merge button?** All of the CI jobs above
--- a/tests/unit/ci/test_required_checks_gate.py
+++ b/tests/unit/ci/test_required_checks_gate.py
@@ -99,3 +99,35 @@ class TestRequiredChecksGate:
             f"{GATE_JOB} must set `if: always()` (got {gate_if!r}) so it does "
             "not skip on label/auto-merge events and deadlock the required check."
         )
+
+    def test_gate_assertion_fires_on_unwired_job(self) -> None:
+        """Negative-path: the invariant check must flag a job absent from needs:.
+
+        Construct a synthetic jobs dict that introduces a new gating job not
+        listed in required-checks-gate.needs and verify the gap is detected.
+        This guards against the invariant check itself silently passing when
+        it should fail — exactly the failure mode that let a red lint job reach
+        main (issue #1315).
+        """
+        # Minimal synthetic workflow jobs: a gate that only needs "job-a",
+        # plus an additional "job-b" that is unwired (not in needs).
+        synthetic_jobs: dict[str, Any] = {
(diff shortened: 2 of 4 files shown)


docs/DEFINITION_OF_DONE.md
```diff
@@ -34,6 +34,12 @@ A piece of work is **done** when every item below is true.
 | 21 | Pre-commit hooks pass on the diff | CI job `pre-commit` |
```
mvillmow: Stale reference: row 21 still cites `CI job pre-commit`, but no `pre-commit` job exists — the pre-commit hook suite was folded into the required `lint` job in #1173 (see `_required.yml:173-179`). This PR edits this exact table (adding the aggregator note directly below row 22), and the implementation plan committed to fixing this row (`pre-commit` → `lint`). Update to: `| 21 | Pre-commit hooks pass on the diff | CI job `lint` (pre-commit suite folded into `lint` per #1173) |`.

tests/unit/ci/test_required_checks_gate.py
```diff
@@ -0,0 +1,101 @@
...
+    """The gate must aggregate every gating job in _required.yml."""
+
+    def test_gate_job_exists(self, jobs: dict[str, Any]) -> None:
+        """_required.yml must define the required-checks-gate job."""
+        assert GATE_JOB in jobs, (
+            f"{GATE_JOB} job missing from _required.yml — it is the single "
+            "branch-protection required check; see docs/ci/required-checks.md"
+        )
+
+    def test_gate_needs_every_non_exempt_job(self, jobs: dict[str, Any]) -> None:
```
mvillmow, line 58: The guard asserts the invariant against the real workflow (good — it fails if a job is left unwired), but there is no test proving the assertion actually *fires* on a gap. The plan specified an extracted `_unwired_jobs(wf, excluded)` helper plus a synthetic-fixture negative test that injects a job absent from `needs:` and asserts it is flagged. Consider adding that negative-path test so a future refactor of the comparison logic can't silently neuter the guard (e.g. feed a fake `jobs` dict with a `dummy-job` not in `needs` and assert it appears in `missing`).

mvillmow (commented):
Correct: lint now structurally gates merges via toJSON(needs) aggregator (injection-safe, improves on plan). Minor: stale DoD row 21 (pre-commit job gone), weaker negative-test than planned. No correctness defects.

PR #1341, opened 2026-06-14
fix(license-scan): classify marker-excluded deps via static fallback

### LINKED ISSUE #1258: Follow-up from #1219: 1 item(s) (core)

mvillmow:
Add `STATIC_FALLBACK_LICENSES` to classify `tomli` and `tzdata` on the Python 3.13/Linux CI leg where their markers exclude them from installation. Previously these deps were skipped with a misleading note; now they resolve to their SPDX license from NOTICE, and unknown marker-excluded deps exit(2) instead of silently passing.

**Changes:**
- `scripts/check_license_compatibility.py`: add `STATIC_FALLBACK_LICENSES` dict (MIT for tomli, Apache-2.0 for tzdata); replace silent-skip branch with fallback-or-exit(2) logic; update docstring with new FAILS LOUDLY bullet
- `tests/unit/scripts/test_check_license_compatibility.py`: rename contradicted test; add `TestStaticFallback` class with 6 tests including two staleness-mitigation cross-checks (NOTICE value bound to per-package lines, real metadata when dep is installed)

Closes #1258

### DIFF
--- a/scripts/check_license_compatibility.py
+++ b/scripts/check_license_compatibility.py
@@ -62,7 +62,7 @@ ALLOWED_EXTRA_COPYLEFT: dict[str, frozenset[str]] = {
 # test suite cross-checks both keys and values against NOTICE and against real
 # installed metadata when the dep is actually installable.
 STATIC_FALLBACK_LICENSES: dict[str, list[str]] = {
-    "tomli": ["MIT"],          # python_version < '3.11'; NOTICE:58
+    "tomli": ["MIT"],  # python_version < '3.11'; NOTICE:58
     "tzdata": ["Apache-2.0"],  # platform_system == 'Windows'; NOTICE:28
 }
 
--- a/scripts/check_license_compatibility.py
+++ b/scripts/check_license_compatibility.py
@@ -15,9 +15,9 @@ FAILS LOUDLY (never silently passes):
     -> coverage hole -> exit 2. CI installs `.[all]` on a setup-python runner first.
   * a license string not in the mapping tables -> exit (PR:1 / main:0) with a
     hint to extend TROVE_TO_SPDX / LICENSE_ALIASES.
-  * a distributed dep whose PEP 508 marker excludes the current interpreter
-    -> classified via FALLBACK_LICENSES (NOTICE-backed); if no entry exists
-    -> exit 2. Add new marker-excluded deps to FALLBACK_LICENSES.
+  * a markered-out dep (installable_now=False) with no entry in
+    STATIC_FALLBACK_LICENSES -> exit 2. Add the package + SPDX id from NOTICE,
+    then update the cross-check tests in tests/unit/scripts/.
 
 Stdlib only (importlib.metadata + packaging, both already runtime deps); runs
 under plain `python3` once the package + extras are pip-installed.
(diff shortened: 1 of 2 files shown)


tests/unit/scripts/test_check_license_compatibility.py
```diff
@@ -257,3 +258,102 @@ def test_pygithub_lgpl_fixture_passes(self, tmp_path):
...
+    """STATIC_FALLBACK_LICENSES classifies marker-excluded deps instead of skipping.
+
+    Staleness mitigations:
+    - test_static_values_match_installed_metadata: cross-checks against real
+      importlib.metadata when the dep IS installable (catches license drift).
+    - test_static_values_match_notice: cross-checks against NOTICE (the
+      authoritative human-readable analysis per the script's own docstring).
+    """
+
+    def test_tomli_fallback_classifies_as_compatible(self):
```
mvillmow, line 273: P4/DRY: `test_tomli_fallback_classifies_as_compatible` is a verbatim duplicate of `test_uninstalled_other_python_dep_with_fallback_classifies_not_fails` (line 143): identical patches, identical assertion, identical package. Since the renamed test in `TestLoudFailure` already covers 'known fallback → compatible, not a hole', this test adds zero new coverage. Either remove it and let the `TestLoudFailure` test carry the contract, or differentiate it (e.g. assert the printed note contains the package name, or assert `ids == ["MIT"]` from a spy).

scripts/check_license_compatibility.py
```diff
@@ -263,13 +277,22 @@ def scan(records: dict[str, dict] | None = None) -> list[tuple[str, list[str]]]:
...
             print(
-                f"note: skipping {pkg!r} — distributed only under another "
-                "Python/platform (marker excludes this interpreter); it is "
-                "classified on the matching CI matrix row.",
+                f"note: {pkg!r} excluded by marker in this env; "
+                f"classifying via static fallback from NOTICE: {fallback}",
                 file=sys.stderr,
             )
-            continue
+            ids = fallback
```
mvillmow, line 296: P7/POLA: The `scan()` docstring (line 264, pre-existing) now contradicts the behavior introduced by this diff. It says marker-excluded deps "are skipped with a note, not treated as a hole" — but lines 280-295 (new in this diff) either classify them via `STATIC_FALLBACK_LICENSES` or `exit(2)`. A caller reading only the docstring will have incorrect expectations. Update the docstring to: "Deps absent only because their marker excludes the current interpreter are classified via `STATIC_FALLBACK_LICENSES`, or exit 2 if the dep has no fallback entry."

mvillmow (commented):
Implementation is correct and complete: fallback dict, exit(2) for unknown deps, and NOTICE cross-check are all sound. Two minor issues: duplicate tomli test and stale scan() docstring.

tests/unit/scripts/test_check_license_compatibility.py
mvillmow: Resolved by the automation reviewer after a GO implementation review.

scripts/check_license_compatibility.py
mvillmow: Resolved by the automation reviewer after a GO implementation review.

PR #1343, opened 2026-06-14
test(ci-gate): extract _unwired_jobs helper so guard and negative test share one code path

### LINKED ISSUE #1338: test(ci-gate): extract _unwired_jobs(wf, excluded) helper so the guard and its negative test share one code path
## Context

Follow-up from a review thread on #1317:
https://github.com/HomericIntelligence/ProjectHephaestus/pull/1317#discussion_r3408395881

> The guard asserts the invariant against the real workflow (good — it fails if a job is left unwired), but there is no test proving the assertion actually *fires* on a gap. The plan specified an extracted `_unwired_jobs(wf, excluded)` helper plus a synthetic-fixture negative test that injects a job absent from `needs:` and asserts it is flagged. Consider adding that negative-path test so a future refactor of the comparison logic can't silently neuter the guard (e.g. feed a fake `jobs` dict with a `dummy-job` not in `needs` and assert it appears in `missing`).

## Current state (partially addressed on #1317)

#1317 added a negative-path test — `tests/unit/ci/test_required_checks_gate.py::test_gate_assertion_fires_on_unwired_job` — which constructs a synthetic `jobs` dict with an unwired `job-b` and asserts the gap is detected. This satisfies the *spirit* of the comment (it proves the assertion fires on a gap).

**However**, the negative test **re-implements** the `missing = expected - gate_needs` comparison *inline* rather than exercising a shared helper. There is no `_unwired_jobs(wf, excluded)` helper (`grep -rn "_unwired_jobs"` returns nothing). So the test and the real guard are two separate copies of the comparison logic — a future refactor of the *real* guard could still silently diverge from what the test checks, which is the exact robustness risk the reviewer flagged.

## Proposed work

1. Extract the unwired-job detection into a pure helper `_unwired_jobs(wf, excluded)` (returns the set of jobs not present in `required-checks-gate.needs`, minus an exclusion set) in the gate module under `hephaestus/ci/`.
2. Refactor the real invariant guard to call `_unwired_jobs(...)`.
3. Refactor `test_gate_assertion_fires_on_unwired_job` (and the positive-path test) to drive the SAME `_unwired_jobs(...)` helper with synthetic fixtures, instead of re-deriving `missing` inline — so both the guard and the test share one code path and cannot diverge.

## Acceptance criteria

- [ ] `_unwired_jobs(wf, excluded)` exists and is the single source of truth for the comparison.
- [ ] The real required-checks-gate guard uses it.
- [ ] The negative-path test injects a synthetic `jobs` dict with a job absent from `needs:` and asserts it is returned by `_unwired_jobs(...)` (not via a re-implemented inline diff).
- [ ] Positive-path test asserts the real workflow yields no unwired jobs via the same helper.
- [ ] `pixi run pytest tests/unit/ci/test_required_checks_gate.py` green.

Follows up #1315 / #1317.

mvillmow:
Extract the unwired-job comparison into a single pure helper `_unwired_jobs(wf, excluded)` in a new leaf module `hephaestus/ci/required_checks_gate.py`, then route both the positive-path guard test and the negative-path synthetic-fixture test through that helper so they cannot diverge on a future refactor.

## Changes

- **New**: `hephaestus/ci/required_checks_gate.py` — contains `GATE_JOB` constant and `_unwired_jobs(wf, excluded, *, gate_job=GATE_JOB)` pure helper (single source of truth for the fan-in comparison)
- **Modified**: `tests/unit/ci/test_required_checks_gate.py` — imports `GATE_JOB` and `_unwired_jobs` from the new module; `test_gate_needs_every_non_exempt_job` now takes `workflow` fixture and calls `_unwired_jobs(workflow, EXEMPT_JOBS)`; `test_gate_assertion_fires_on_unwired_job` builds a workflow-shaped synthetic dict and calls the same helper instead of re-deriving `missing` inline

Closes #1338

### DIFF
--- a/hephaestus/ci/required_checks_gate.py
+++ b/hephaestus/ci/required_checks_gate.py
@@ -1,0 +1,45 @@
+"""Single source of truth for the required-checks-gate fan-in invariant.
+
+The ``required-checks-gate`` job in ``.github/workflows/_required.yml`` is the
+only branch-protection required status check for that workflow. It must fan in
+every gating job via its ``needs:`` list; a job left out silently stops
+blocking merges (issue #1315). ``_unwired_jobs`` computes which jobs are *not*
+wired into the gate, and is shared by the structural guard tests so the guard
+and its negative-path test exercise one code path and cannot diverge
+(issue #1338).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+from typing import Any
+
+GATE_JOB = "required-checks-gate"
+
+
+def _unwired_jobs(
+    wf: dict[str, Any],
+    excluded: Iterable[str],
+    *,
+    gate_job: str = GATE_JOB,
+) -> set[str]:
+    """Return jobs defined in ``wf`` but absent from the gate's ``needs:`` list.
+
+    Args:
+        wf: A parsed GitHub Actions workflow document (the mapping produced by
+            ``yaml.safe_load``). Must contain a ``jobs`` mapping that includes
+            ``gate_job`` with a ``needs`` list.
+        excluded: Job names intentionally not gated (e.g. advisory jobs and the
+            gate itself); these are removed from the result.
+        gate_job: Name of the aggregating gate job. Defaults to
+            ``required-checks-gate``.
+
+    Returns:
+        The set of job names that are defined in ``wf['jobs']`` but neither
+        listed in ``wf['jobs'][gate_job]['needs']`` nor in ``excluded``. An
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
LGTM. `_unwired_jobs` correctly extracted as single source of truth; both guard test and negative test now call the shared helper; `GATE_JOB` constant de-duplicated; `__init__.py` correctly left unchanged (private helper stays un-exported).

Issue #1289, opened 2026-06-13
[refactor] Decompose CIDriver god-class into 4 narrow collaborators (SRP)


mvillmow:
## Summary

`CIDriver` was a 3,338-line, 50-method god-class violating SRP. This issue tracks extracting four narrow collaborators using the Dependency Inversion Principle (DIP) pattern.

## Collaborators

| Module | Responsibility |
|--------|---------------|
| `pr_discovery.py` | Viewer-login caching + PR enumeration (issue-driven, bot, failing) |
| `ci_check_inspector.py` | Required CI check queries (`gh pr checks`) |
| `ci_fix_orchestrator.py` | CI fix session management (codex + claude paths) |
| `post_merge_processor.py` | Drive-green `/learn` + `/compact` after PR merges |

## Target

Reduce `ci_driver.py` from 3,338 → ≤2,450 lines (delegation stubs only).

## Design Constraints

- Narrow-callable injection (DIP): collaborators receive `Callable[[], T]` providers, never `self`/`CIDriver`
- All existing `patch.object(CIDriver, '_method')` test targets preserved via delegation stubs
- Lambda wrappers for all injected callables so `patch.object` continues to work in tests

Refs #1179

mvillmow:
# Implementation Plan


Now I have all signatures. Key findings:
- `_resolve_dirty_pr`: **3 params** `(self, issue_number, pr_number, acquired_slot) -> WorkerResult` — no `worktree_path`
- `_arm_drive_green`: `(self, pr_number, pr_head_branch, pr_head_sha) -> None` — confirmed
- `_sweep_orphaned_arming_records`: `(self) -> None` — confirmed
- `_mark_drive_green_learn_result`: keyword-only! `(self, issue_number, record: dict[str, Any], *, succeeded: bool) -> None`
- `_wait_for_pr_terminal`: `(self, issue_number, pr_number) -> str` — confirmed
- `_attempt_mechanical_rebase`: `(self, issue_number, pr_number, acquired_slot) -> bool` — confirmed (has `acquired_slot`)

The plan also had `_mark_drive_green_learn_result(self, issue_number, pr_number, record)` — wrong: it's `(self, issue_number, record, *, succeeded: bool)` — no `pr_number`, keyword-only `succeeded`.

# Implementation Plan

## Objective

[...]

PR #1346, opened 2026-06-14
test(automation): regression-guard FAILING_CHECK_CONCLUSIONS / _pr_is_failing as single definitions

### LINKED ISSUE #1345: Follow-up from #1289: 2 item(s) (core, safety)
_Consolidated follow-up from implementation of #1289._

### CICheckInspector None-slot calls crash with TypeError when unwired

Seven slots on `CICheckInspector` are initialized to `None` (`ci_check_inspector.py:57-64`) but called without a None guard in `_sweep_orphaned_arming_records` (line 301: `self._load_arming_state_fn(issue_number)`) and `_check_arming_on_drive_start` (line 359: same). Constructing `CICheckInspector` and calling either method without wiring all slots raises `TypeError: 'NoneType' object is not callable` — a hard crash with no informative message. Repro: `ins = CICheckInspector(options=..., ...); ins._check_arming_on_drive_start(1, 1)`. Sketch fix: replace `None` defaults with lambda stubs that raise `NotImplementedError('slot not wired: _load_arming_state_fn')`, mirroring the pattern used for `PRDiscovery._unwired_*` methods added in this PR.

### Deduplicate FAILING_CHECK_CONCLUSIONS and _pr_is_failing across 3 modules

The extraction created three copies of the same frozenset and predicate:
- `ci_driver.py:58` — `FAILING_CHECK_CONCLUSIONS`
- `post_merge_processor.py:35` — `FAILING_CHECK_CONCLUSIONS` (comment says 'keep in sync')
- `pr_discovery.py:26` — `_FAILING_CHECK_CONCLUSIONS` (module-private copy)

Similarly `_pr_is_failing` is defined at `ci_driver.py:61`, `post_merge_processor.py:38`, and `pr_discovery.py:29` (as `_pr_is_failing_local`). A single canonical definition in a new leaf module (e.g. `hephaestus/automation/pr_predicates.py`) imported by all three would eliminate the drift risk. Sketch fix: create `pr_predicates.py` exporting `FAILING_CHECK_CONCLUSIONS: frozenset[str]` and `pr_is_failing(pr) -> bool`; replace all three copies with `from .pr_predicates import FAILING_CHECK_CONCLUSIONS, pr_is_failing`.

---

mvillmow:
## Summary

Issue #1345 describes two follow-up items from a PR (#1289) that was never merged:

1. **Safety**: Guard `CICheckInspector` None-slot callable fields — the class does not exist anywhere in the repo.
2. **Core library**: Deduplicate `FAILING_CHECK_CONCLUSIONS` / `_pr_is_failing` across three modules — only one copy has ever existed (in `ci_driver.py`); the three-copy state described in the issue was never introduced.

Every concrete premise of the issue is factually false against `main`. The correct action is not to fabricate the modules the issue presupposes, but to lock in the already-satisfied DRY invariant as a machine-checkable regression guard.

This PR adds `TestFailingCheckPredicateSingleDefinition` to the failing-PR discovery test suite. The two new tests use `ast.parse` to scan every `.py` file under `hephaestus/automation/` and assert:

- `FAILING_CHECK_CONCLUSIONS` is assigned exactly once (in `ci_driver.py`)
- `_pr_is_failing` is defined exactly once (in `ci_driver.py`)

If a future PR re-introduces duplicate copies, CI will catch it immediately.

## Verification evidence

```
find hephaestus -name ci_check_inspector.py -o -name pr_discovery.py -o -name post_merge_processor.py
# → no output (modules do not exist)

grep -rln "CICheckInspector" hephaestus tests --include="*.py"
# → no output (class does not exist)

grep -rn "FAILING_CHECK_CONCLUSIONS\|def _pr_is_failing" hephaestus/
# → exactly: ci_driver.py:79 and ci_driver.py:82
```

## Test plan

- [x] `pixi run pytest tests/unit/automation/test_ci_driver_failing_pr_discovery.py -v` — all 25 tests pass (2 new)
- [x] `pixi run pytest tests/unit/automation -q` — 1659 tests pass, no regressions

Closes #1345

### DIFF
--- a/tests/unit/automation/test_ci_driver_failing_pr_discovery.py
+++ b/tests/unit/automation/test_ci_driver_failing_pr_discovery.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import ast
 import json
+from pathlib import Path
 from typing import Any
 from unittest.mock import Mock, patch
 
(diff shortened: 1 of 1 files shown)


tests/unit/automation/test_ci_driver_failing_pr_discovery.py
```diff
@@ -317,3 +319,68 @@ def test_is_bot_pr_mode_holds_for_synthetic_failing_pr_key(self, ci_driver: CIDr
...
+    def _count_function_defs(self, name: str) -> list[Path]:
+        """Return paths of automation modules that define a function named *name*."""
+        hits: list[Path] = []
+        for py_file in self._AUTOMATION_DIR.rglob("*.py"):
+            try:
+                tree = ast.parse(py_file.read_text(encoding="utf-8"))
+            except SyntaxError:
+                continue
+            for node in ast.walk(tree):
+                if isinstance(node, ast.FunctionDef) and node.name == name:
```
mvillmow, line 362: `_count_function_defs` only checks `ast.FunctionDef`, missing `ast.AsyncFunctionDef`. If `_pr_is_failing` were ever redefined as `async def`, the guard would return zero hits and the assertion `len(hits) == 1` would fail with 'found 0' — effectively a false negative that makes the guard invisible. Add `or isinstance(node, ast.AsyncFunctionDef)` to the condition to close this gap.

mvillmow (commented):
Regression-guard tests are correct and proportionate. One minor gap: AsyncFunctionDef not covered in _count_function_defs, which could silently miss an async redefinition.

tests/unit/automation/test_ci_driver_failing_pr_discovery.py
mvillmow: Resolved by the automation reviewer after a GO implementation review.

PR #1595, opened 2026-06-24
feat(scripts): add sanitized Pi smoke validation

### LINKED ISSUE #1580: [Feature]: Add sanitized Pi smoke validation for local and Slurm execution
### Motivation

Operators need a repeatable way to verify Pi against private OpenAI-compatible model aliases, including longer runs launched from the internal execution environment, without storing sensitive details in the repository.

### Proposed Solution

Add a smoke harness that accepts provider and model aliases only through environment variables and runs read-only prompts through Pi. Add a Slurm wrapper template with placeholders only. Validation logs stay local and untracked. Tests should fake Pi output and fake Slurm submission commands to verify command construction without requiring the internal environment.

### Alternatives Considered

Embedding concrete internal endpoints in smoke scripts was rejected as a leak risk. Making live private endpoint checks part of public CI was rejected because CI cannot access the internal environment and must not contain private configuration.

### Severity

major

### Parent Epic (optional)

_No response_

mvillmow:
## Summary
Adds sanitized Pi smoke validation for local and Slurm execution without committing private provider details or validation logs.

## Changes
- Added Pi smoke and Slurm wrapper scripts that take provider/model aliases from environment variables and use placeholder-only Slurm configuration.
- Updated ignore rules, script documentation, and private denylist checks to keep local smoke logs and sensitive endpoint details out of the repository.
- Adjusted related agent runtime and automation tests to match the updated invocation behavior.

## Testing
- Added and updated unit tests that fake Pi output and Slurm submission commands to verify smoke command construction without requiring the internal environment.

## Closes
Closes #1580


mvillmow (commented):
NOGO: `_run_pi_command` at line 720 calls `_pi_env()` without `model=model`, so HEPH_PI_MODEL is never set in the Pi subprocess env. Lines 724/733 use `_model_from_pi_cmd(cmd)` which now always returns '' — model alias is missing from error/timeout redaction tokens.

scripts/pi_smoke.py
```diff
@@ -32,24 +38,59 @@ def build_parser() -> argparse.ArgumentParser:
...
-    if not model:
-        print("ERROR: HEPH_PI_MODEL must name an operator-local Pi model alias.", file=sys.stderr)
+    missing = _missing_alias_env()
+    if missing:
+        print(f"ERROR: missing required env vars: {', '.join(missing)}", file=sys.stderr)
         return 2
+    model = os.environ.get(PI_MODEL_ENV, "").strip()
     redaction_tokens = pi_private_redaction_tokens(args.cwd, model)
     try:
         result = run_pi_session(
```
mvillmow, line 92: `run_pi_session()` is called without `model=model`. Line 87 reads the alias from `PI_MODEL_ENV` into `model`, but it is never forwarded. Since `_pi_env(model=model)` in the runtime is the intended propagation path, omitting `model=model` here means the runtime's `_pi_env` receives an empty string. Fix: add `model=model` to the `run_pi_session(...)` call.

scripts/pi_smoke_slurm.py
```diff
@@ -0,0 +1,79 @@
...
+def build_parser() -> argparse.ArgumentParser:
+    """Build the Pi Slurm smoke submission parser."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--log-dir", type=Path, default=DEFAULT_LOG_DIR)
+    parser.add_argument("--template", type=Path, default=DEFAULT_TEMPLATE)
+    parser.add_argument("--sbatch", default="sbatch")
+    return parser
+
+
+def _missing_alias_env() -> list[str]:
```
mvillmow: `_missing_alias_env()` and `REQUIRED_ALIAS_ENVS` are byte-for-byte duplicates of `scripts/pi_smoke.py:50-52` and `scripts/pi_smoke.py:32`. Both scripts already import from `hephaestus.agents.runtime`; move the shared constant and helper there (or a shared private utility) to avoid divergence.

mvillmow (commented):
NOGO: two critical bugs — runtime.py:720 calls _pi_env() without model (alias never reaches Pi subprocess env), and pi_smoke.py:90 omits model=model from run_pi_session() (alias silently dropped). Core privacy guarantee is undelivered.

scripts/pi_smoke_slurm.py
```diff
@@ -0,0 +1,73 @@
...
+
+def build_sbatch_cmd(args: argparse.Namespace) -> list[str]:
+    """Build an sbatch command that exports alias env var names only."""
+    log_dir: Path = args.log_dir
+    log_dir.mkdir(parents=True, exist_ok=True)
+    return [
+        args.sbatch,
+        f"--export={','.join(EXPORT_NAMES)}",
+        f"--output={log_dir / 'pi-smoke-%j.out'}",
+        f"--error={log_dir / 'pi-smoke-%j.err'}",
```
mvillmow, line 36: `build_sbatch_cmd` calls `log_dir.mkdir(parents=True, exist_ok=True)` as a side effect of a function named `build_*`. A builder should return a value without mutating the filesystem; callers cannot call `build_sbatch_cmd` (e.g. in tests or dry-run paths) without triggering the mkdir. Move the `mkdir` call to `main()` before `build_sbatch_cmd` is called.

scripts/pi_smoke.py
```diff
@@ -32,25 +38,56 @@ def build_parser() -> argparse.ArgumentParser:
...
+            redacted = redacted.replace(value, f"<redacted:{name}>")
+    return redacted
+
+
+def _write_smoke_log(log_dir: Path, result: AgentRunResult) -> Path:
+    """Write the local Pi smoke result log and return its path."""
+    log_dir.mkdir(parents=True, exist_ok=True)
+    log_path = log_dir / "pi-smoke-local.log"
+    lines = [
+        f"session_id: {result.session_id or ''}",
```
mvillmow, line 72: `_write_smoke_log` always writes to `log_dir / 'pi-smoke-local.log'` (fixed name). Repeated smoke runs silently overwrite the previous log with no warning. Consider appending a timestamp suffix (e.g. `pi-smoke-local-<epoch>.log`) so consecutive runs produce distinct artifacts, or at minimum print a note that the prior log was overwritten.

mvillmow (commented):
Both prior review threads addressed; argv alias leak fixed via env propagation; DRY enforced via runtime exports. Two minor POLA issues: build_sbatch_cmd has mkdir side effect; log file silently overwrites on repeated runs.

tests/unit/scripts/test_pi_smoke.py
```diff
@@ -18,31 +18,62 @@
...
+    assert "LOG_FILE=" in captured.err
+    log_path = Path(captured.err.strip().split("LOG_FILE=", 1)[1])
+    log_text = log_path.read_text(encoding="utf-8")
+    assert "stdout: OK" in log_text
+    assert "private-provider-alias" not in log_text
+    assert "private-model-alias" not in log_text
 
 
 def test_failure_output_redacts_private_values(
     monkeypatch: pytest.MonkeyPatch, tmp_path: Path, capsys: pytest.CaptureFixture[str]
```
mvillmow, line 74: This parse is broken when `session_id` is non-empty. The fixture sets `session_id='pi-smoke'`, so stderr is `'NOTE: ...\nLOG_FILE=<path>\nSESSION_ID=pi-smoke'`. After `.strip()`, splitting on `'LOG_FILE='` gives `'<path>\nSESSION_ID=pi-smoke'` — `Path(...)` of that string includes the newline and trailing line, causing `log_path.read_text()` to raise `FileNotFoundError`. Fix: parse only the LOG_FILE line, e.g. `next(l.split('LOG_FILE=', 1)[1] for l in captured.err.splitlines() if l.startswith('LOG_FILE='))`.

tests/unit/scripts/test_pi_smoke.py
```diff
@@ -18,31 +18,62 @@
...
 
     assert _mod.main([]) == 2
 
 
-def test_runs_pi_with_env_model_alias(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
-    """The smoke harness forwards only the operator-provided model alias."""
-    monkeypatch.setenv("HEPH_PI_MODEL", "private-test-alias")
+def test_runs_pi_with_env_aliases_without_alias_kwargs(
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
```
mvillmow, line 33: Test name `test_runs_pi_with_env_aliases_without_alias_kwargs` is misleading: `model` IS still passed as a kwarg to `run_pi_session` (line 73 asserts `kwargs['model'] == 'private-model-alias'`). The 'without_alias_kwargs' framing implies no alias kwargs at all. Rename to e.g. `test_runs_pi_with_env_aliases_model_via_kwarg_not_argv` to match what is actually asserted.

mvillmow (commented):
Critical test bug: LOG_FILE path parse includes SESSION_ID trailing line when session_id is non-empty, causing FileNotFoundError on log_path.read_text(). Fix the line-extraction in test_pi_smoke.py:73.

mvillmow (commented):
Argv-leak fix and provider env validation are correct. Two minor test gaps: success-redact test writes files to cwd instead of tmp_path; resume_pi_session test omits env propagation assertion that run_pi_session test includes.

tests/unit/agents/test_runtime.py
```diff
@@ -655,10 +652,13 @@ def fake_run(cmd: list[str], **kwargs: Any) -> subprocess.CompletedProcess[str]:
             "private feedback content",
             cwd=tmp_path,
             timeout=30,
+            model="private-alias",
         )
 
     assert captured["cmd"][:-1] == ["pi", "--mode", "json", "--session", "pi-session-789"]
     assert captured["cmd"][-1].startswith("@")
+    assert "--model" not in captured["cmd"]
```
mvillmow, line 660: `test_resume_pi_session_passes_resume_id_without_alias_argv_leak` asserts the alias is absent from argv but does not assert `captured['kwargs']['env']['HEPH_PI_MODEL'] == 'private-alias'`. The sibling `run_pi_session` test (line 513) does. Add the env assertion here so both code paths are symmetrically verified.

mvillmow (commented):
Argv-leak fix is correct: model alias moves to child env only, never argv. One dead-code function (_model_from_pi_cmd) left over from the refactor — harmless due to env fallback, but misleading.

PR #1618, opened 2026-06-26
refactor(automation): extract shared worker options base

### LINKED ISSUE #1388: [refactor] Extract shared parallel-worker execution loop (6 duplicated ThreadPoolExecutor patterns)
## Motivation

`hephaestus/automation/models.py` defines 6 Options classes (`PlannerOptions`, `ImplementerOptions`, `ReviewerOptions`, `PlanReviewerOptions`, `AddressReviewOptions`, `CIDriverOptions`) that share common fields: `max_workers: int = 3` (or `parallel: int = 3`), `state_dir`, `dry_run`, `verbose`. Each re-declares these with identical defaults.

## Proposed Solution

Extract a `WorkerOptionsBase(BaseModel)` with the shared fields. Each Options class inherits from it and adds only its unique fields. This reduces duplication and ensures defaults stay in sync.

## Impact

- ~30 lines removed
- Single place to change default `max_workers` or add a new shared field
- Low risk — Pydantic models compose well with inheritance

## Files affected

- `hephaestus/automation/models.py` (add base class, update 6 Options classes)
- All files that construct Options objects (update field imports if needed)

mvillmow:
## Summary
Extract shared automation worker option fields into a common base model and update worker-facing automation paths to use the unified contract.

## Changes
- Added a shared worker options base in `hephaestus/automation/models.py` for common fields such as worker count, state directory, dry run, and verbosity.
- Updated address review, CI driver, PR reviewer, and review utility code to use the shared options structure.
- Added unit coverage for the options contract and removed obsolete review utility option tests.

## Testing
- Not run; no test command was provided in the supplied issue, changed files, or commits.

## Closes
Closes #1388

### DIFF
--- a/hephaestus/automation/models.py
+++ b/hephaestus/automation/models.py
@@ -142,7 +142,28 @@ class PlanResult(BaseModel):
     plan_already_exists: bool = False
 
 
-class PlannerOptions(BaseModel):
+DEFAULT_WORKER_COUNT = 3
+
+
+class WorkerOptionsBase(BaseModel):
+    """Shared options for automation worker stages."""
+
+    dry_run: bool = False
+
+
+class ParallelWorkerOptionsBase(WorkerOptionsBase):
+    """Shared options for stages that expose ``max_workers``."""
+
+    max_workers: int = DEFAULT_WORKER_COUNT
+
+
+class VerboseParallelWorkerOptionsBase(ParallelWorkerOptionsBase):
+    """Shared worker options for stages with verbose logging."""
+
+    verbose: bool = False
+
+
+class PlannerOptions(WorkerOptionsBase):
     """Options for the Planner."""
 
     issues: list[int]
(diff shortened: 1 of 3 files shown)


hephaestus/automation/models.py
```diff
@@ -142,7 +142,28 @@ class PlanResult(BaseModel):
     plan_already_exists: bool = False
 
 
-class PlannerOptions(BaseModel):
+DEFAULT_WORKER_COUNT = 3
```
mvillmow, line 145: `DEFAULT_WORKER_COUNT` currently centralizes only the Pydantic model defaults. The normal CLI paths pass argparse defaults into these option models, so changing this constant later would not update omitted `--parallel`/`--max-workers` behavior. Please either reuse this constant in the parser defaults too, or add parser-default contract tests that would catch drift.

---

`DEFAULT_WORKER_COUNT` is only used by the Pydantic model defaults, but the actual CLI paths still hard-code `3` in `planner._build_parser` (`--parallel`) and `_review_utils.add_max_workers_arg` (`--max-workers`). Since each `main()` passes the parsed argparse value into the option model, changing this constant will not change the default behavior users get from the CLIs. Import/use this constant in the parser defaults/help text and add a parser-default assertion so the worker-count default is genuinely single-sourced.

mvillmow (commented):
GO with one minor default-drift maintainability note.

tests/unit/automation/test_options_contract.py
```diff
@@ -0,0 +1,219 @@
...
+from __future__ import annotations
+
+import inspect
+from collections.abc import Callable
+from typing import Any
+
+import pytest
+from pydantic import BaseModel
+
+import hephaestus.automation.models as automation_models
```
mvillmow, line 12: Re-opening: prior review comment not addressed — The current diff still has both import styles in tests/unit/automation/test_options_contract.py and continues using automation_models.* references instead of direct imports.

> Module 'hephaestus.automation.models' is imported with both 'import' and 'import from'.

mvillmow (commented):

mvillmow (commented):
LGTM

Issue #1554, opened 2026-06-20
[audit][nitpick] Packaging/DX gaps: MIGRATION link, OS CI, recursive audit task, no scaffolding


mvillmow:
- **[S12 Packaging]** MIGRATION.md not linked from README.md — `README.md`
- **[S12 Packaging]** macOS/Windows CI coverage is documented-absent (tracked issue #539) — `.github/workflows/test.yml:7-12`
- **[S13 Developer Experience]** `pixi.toml` [tasks] `audit` task uses `pixi run --environment lint` inside a pixi task (recursive invocation) — `pixi.toml:18`
- **[S13 Developer Experience]** No codegen/scaffolding tooling for new subpackages or new CLI scripts — `scripts/`

Split from #1518 (NITPICK audit bundle, strict full-coverage audit 2026-06-16). Low priority; address opportunistically.

mvillmow:
# Implementation Plan


I have all the evidence I need. Let me make a scoping decision on item 4 (scaffolding), since it's the most open-ended and the issue is explicitly low-priority/nitpick. I'll scope it to a single, focused subpackage scaffolder that follows the existing `scripts/` shim + `hephaestus.<pkg>` pattern, with the option to also scaffold a CLI script. Now I'll write the plan.

# Implementation Plan

## Objective

Issue #1554 bundles four low-priority packaging/DX nitpicks split from #1518. Address all four: (1) link the existing `docs/MIGRATION.md` from `README.md` (it exists but is unreferenced — only `COMPATIBILITY.md` and `docs/DEFINITION_OF_DONE.md` link it); (2) document macOS/Windows CI absence — already substantially done in `.github/workflows/test.yml:1-13` header but verify/tighten and add the same scope note to `_required.yml`; (3) fix the recursive pixi invocation in `[tasks] audit` at `pixi.toml:18` (`pixi run audit` spawns a nested `pixi run --environment lint pip-audit`); (4) add a minimal subpackage/CLI scaffolding script under `scripts/` following the existing shim + `hephaestus.scripts_lib` pattern. Net effect: cleaner docs cross-linking, a non-recursive audit task, and a codegen tool that emits boilerplate matching repo conventions.

## Approach

[...]

PR #1744, opened 2026-07-01
fix(automation): enforce signed DCO commits

### LINKED ISSUE #1491: [audit][S10 Planning] DCO (Signed-off-by) requirement is documented but not mechanically enforced

**Section:** 10 — Planning
**Severity:** MINOR
**Location:** `CONTRIBUTING.md:219-240`
**Principle:** KISS/POLA — stated requirements that go unenforced create confusion about what is actually required.

## Detail

CONTRIBUTING.md documents a Developer Certificate of Origin requirement and instructs contributors to use 'git commit -s -S'. However: (1) no pre-commit hook validates Signed-off-by trailers, (2) the pr-policy CI gate only verifies cryptographic signatures (-S), not DCO signoff (-s), and (3) git log of the 10 most recent commits shows zero Signed-off-by trailers. The requirement is stated as mandatory ('both are required') but is unenforced in practice.

mvillmow:
## Summary
- Enforce signed and DCO-signed commits in automation-created commits and mechanical rebases.
- Update CI/prompt remediation text to consistently point operators at git commit -S -s.
- Add unit coverage for commit metadata repair before push and loop rebase behavior.

## Test Plan
- [x] pixi run pytest tests/unit/automation/test_git_utils.py tests/unit/automation/test_loop_repo_manager.py tests/unit/automation/test_pr_manager.py tests/unit/scripts/test_check_dco_signoff.py -q --no-cov
- [x] pixi run ruff check hephaestus/automation/ci_fix_orchestrator.py hephaestus/automation/git_utils.py hephaestus/automation/prompts/implementation.py scripts/check_dco_signoff.py tests/unit/automation/test_ci_driver.py tests/unit/automation/test_git_utils.py tests/unit/automation/test_implementer_phase_runner.py tests/unit/automation/test_loop_repo_manager.py tests/unit/scripts/test_check_dco_signoff.py

Closes #1491

Signed-off-by: Micah Villmow <4211002+mvillmow@users.noreply.github.com>

### DIFF
--- a/.github/workflows/_required.yml
+++ b/.github/workflows/_required.yml
@@ -420,7 +420,7 @@ jobs:
             # shellcheck disable=SC2086  # word-splitting of $bad is intentional
             printf '  %s\n' $bad
             msg="::error::Every commit MUST be cryptographically signed."
-            msg+=" Re-sign with: git rebase --exec 'git commit --amend --no-edit -S' origin/main"
+            msg+=" Re-sign with: git rebase --exec 'git commit --amend --no-edit -S -s' origin/main"
             echo "$msg"
             exit 1
           fi
--- a/hephaestus/automation/ci_driver.py
+++ b/hephaestus/automation/ci_driver.py
@@ -1114,6 +1114,7 @@ class _CIDriverCore:
             issue_number=issue_number,
             pr_number=pr_number,
             pr_head_branch=pr_head_branch,
+            pr_base_branch=self._get_pr_base_branch(pr_number),
             session_id=None,
         )
         if not pushed:
@@ -1192,6 +1193,7 @@ class _CIDriverCore:
             # CI fix push must target THIS remote ref even if Claude switches
             # branches locally during the session (#832).
             pr_head_branch = self._get_pr_branch(pr_number)
+            pr_base_branch = self._get_pr_base_branch(pr_number)
 
             if self.options.dry_run:
                 logger.info(
@@ -1214,6 +1216,7 @@ class _CIDriverCore:
                 session_id,
                 advise_findings,
                 pr_head_branch=pr_head_branch,
+                pr_base_branch=pr_base_branch,
             )
             if fixed:
                 logger.info(
@@ -1322,6 +1325,16 @@ class _CIDriverCore:
         except Exception as e:
             logger.warning("Could not fetch branch for PR #%s: %s", pr_number, e)
             return f"pr-{pr_number}"
+
+    def _get_pr_base_branch(self, pr_number: int) -> str:
+        """Return the PR base branch, defaulting to ``main`` on lookup gaps."""
+        try:
+            gh_state = self._gh_pr_state(pr_number) or {}
+        except Exception as exc:
+            logger.warning("Could not fetch base branch for PR #%s: %s", pr_number, exc)
+            return "main"
+        base_branch = str(gh_state.get("baseRefName") or "main")
+        return base_branch or "main"
 
     def _get_worktree_path(self, issue_number: int, pr_number: int) -> Path:
         """Resolve the worktree path for a given issue/PR.
@@ -1433,7 +1446,7 @@ class _CIDriverCore:
             )
 
     def _gh_pr_state(self, pr_number: int) -> dict[str, Any] | None:
-        """Return ``{state, headRefOid, mergedAt, mergeStateStatus}`` or ``None``.
+        """Return ``{state, headRefOid, mergedAt, mergeStateStatus, baseRefName}`` or ``None``.
 
         Used by the on-drive-start check (#840) to detect post-merge so the
         ``/learn`` capture can fire exactly once per issue, and by
@@ -1447,7 +1460,7 @@ class _CIDriverCore:
                     "view",
                     str(pr_number),
                     "--json",
-                    "state,headRefOid,mergedAt,mergeStateStatus",
+                    "state,headRefOid,mergedAt,mergeStateStatus,baseRefName",
                 ],
                 check=False,
             )
@@ -2000,7 +2013,8 @@ class _CIDriverCore:
         issue_number: int,
         pr_number: int,
         pr_head_branch: str,
-        session_id: str | None,
+        session_id: str | None = None,
+        pr_base_branch: str = "main",
     ) -> bool:
         """Delegate to CIFixOrchestrator (extracted #1357)."""
         return self._fix_orchestrator.push_ci_fix(
@@ -2009,6 +2023,7 @@ class _CIDriverCore:
             issue_number=issue_number,
             pr_number=pr_number,
             pr_head_branch=pr_head_branch,
+            pr_base_branch=pr_base_branch,
             session_id=session_id,
         )
 
@@ -2104,6 +2119,7 @@ class _CIDriverCore:
         advise_findings: str = "",
         *,
         pr_head_branch: str,
+        pr_base_branch: str = "main",
     ) -> bool:
         """Delegate to CIFixOrchestrator (extracted #1357)."""
         return self._fix_orchestrator.run_ci_fix_session(
@@ -2114,6 +2130,7 @@ class _CIDriverCore:
             session_id,
             advise_findings,
             pr_head_branch=pr_head_branch,
+            pr_base_branch=pr_base_branch,
         )
 
     def _enable_auto_merge(self, pr_number: int, is_bot_pr: bool = False) -> bool:
--- a/hephaestus/automation/ci_fix_orchestrator.py
+++ b/hephaestus/automation/ci_fix_orchestrator.py
@@ -566,12 +566,8 @@ class CIFixOrchestrator:
             worktree_path: Path to the checked-out worktree.
             ci_logs: Combined CI failure log text.
             session_id: Optional agent session ID to resume.
-            advise_findings: Prior learnings from the advise step, prepended to
-                the prompt as context. Empty or a skip marker contributes nothing.
-            pr_head_branch: The PR's head-branch name on the remote. The push
-                uses this as the destination refspec so the fix lands on the
-                actual PR branch even if the agent switched branches locally
-                during the session (#832).
+            advise_findings: Prior learnings prepended to the prompt.
+            pr_head_branch: Remote PR head branch used as the push destination.
 
         Returns:
             True if the fix session succeeded and the branch was pushed.
--- a/hephaestus/automation/ci_fix_orchestrator.py
+++ b/hephaestus/automation/ci_fix_orchestrator.py
@@ -567,10 +567,9 @@ class CIFixOrchestrator:
             ci_logs: Combined CI failure log text.
             session_id: Optional agent session ID to resume.
             advise_findings: Prior learnings prepended to the prompt.
-            pr_head_branch: Remote PR head branch used as the push destination.
-                Threaded explicitly (not read from the local worktree) because
-                an agent can switch branches mid-session — the push must target
-                the PR's original remote refspec regardless (#832).
+            pr_head_branch: Remote PR head branch the push targets, even if
+                the agent switches branches mid-session (#832).
+            pr_base_branch: PR base branch used for commit-metadata repair.
 
         Returns:
             True if the fix session succeeded and the branch was pushed.
--- a/hephaestus/automation/ci_fix_orchestrator.py
+++ b/hephaestus/automation/ci_fix_orchestrator.py
@@ -292,6 +292,7 @@ class CIFixOrchestrator:
         pr_number: int,
         pr_head_branch: str,
         session_id: str | None,
+        pr_base_branch: str = "main",
     ) -> bool:
         """Check head advancement, retry if needed, then push CI fixes.
 
@@ -308,10 +309,12 @@ class CIFixOrchestrator:
                 session_id=session_id,
             ):
                 return False
-        if not self._ci_fix_head_is_pushable(worktree_path, issue_number):
+        base_ref = f"origin/{pr_base_branch}"
+        if not self._ci_fix_head_is_pushable(worktree_path, issue_number, base_ref=base_ref):
             if not self._ci_fix_residual_commit_is_safe(
                 worktree_path=worktree_path,
                 issue_number=issue_number,
+                base_ref=base_ref,
             ):
                 return False
             if not self._commit_residual_ci_fix_changes(
@@ -319,16 +322,18 @@ class CIFixOrchestrator:
                 issue_number=issue_number,
             ):
                 return False
-            if not self._ci_fix_head_is_pushable(worktree_path, issue_number):
+            if not self._ci_fix_head_is_pushable(worktree_path, issue_number, base_ref=base_ref):
                 return False
         try:
-            ensure_branch_commit_metadata(worktree_path)
+            ensure_branch_commit_metadata(worktree_path, base_branch=pr_base_branch)
         except Exception as metadata_err:
             logger.error(
                 "Issue #%s: failed to enforce signed+DCO commit metadata before push: %s",
                 issue_number,
                 metadata_err,
             )
+            return False
+        if not self._ci_fix_head_is_pushable(worktree_path, issue_number, base_ref=base_ref):
             return False
         try:
             push_current_branch_with_lease_on_divergence(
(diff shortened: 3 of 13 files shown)


hephaestus/automation/git_utils.py
```diff
@@ -535,6 +535,23 @@ def _remove_untracked_files_tracked_by_ref(cwd: Path, ref: str) -> list[Path]:
...
+def ensure_branch_commit_metadata(
+    cwd: Path,
+    base_branch: str = "main",
+    *,
+    remote: str = "origin",
+) -> None:
+    """Rewrite branch commits so each carries a verified signature and DCO trailer."""
+    base_ref = f"{remote}/{base_branch}"
+    run(["git", "fetch", remote, base_branch], cwd=cwd)
+    run(_commit_policy_rebase_command(base_ref), cwd=cwd)
```
mvillmow: This helper runs `git rebase ... --exec ...` but does not abort on conflict. `push_ci_fix()` catches the resulting exception and returns `False`, leaving the worktree in an in-progress rebase state for subsequent automation. Match `rebase_worktree_onto()` here: catch `subprocess.CalledProcessError`, run `git rebase --abort` with `check=False`, then re-raise or return a failure signal; add a unit test for the conflict path.

---

If this rebase conflicts or an exec amend fails after rebase state is created, the caller catches the exception and returns `False`, but the worktree is left mid-rebase. That poisons the next automation attempt. Mirror `rebase_worktree_onto()` here: catch `CalledProcessError`, run `git rebase --abort` best-effort, then re-raise or return a failure signal.

---

Re-opening: prior review comment not addressed — `ensure_branch_commit_metadata()` still has no `CalledProcessError` handler or best-effort `git rebase --abort`; `_push_ci_fix()` catches the exception and returns `False`, so a failed rebase can still leave rebase state behind.

> This helper runs `git rebase ... --exec ...` but does not abort on conflict.

---

If this `git rebase ... --exec` fails, the helper raises without aborting the rebase. `push_ci_fix()` catches the exception and returns `False`, leaving the automation worktree in an in-progress/conflicted rebase; the existing `rebase_worktree_onto()` path aborts specifically to avoid that state. Wrap this rebase in `try/except` and run `git rebase --abort` with `check=False`, or reuse the existing helper's cleanup semantics.

mvillmow (commented):
NOGO: metadata rebase needs conflict cleanup before merge.

mvillmow (commented):
NOGO: pre-push metadata rewrite is not reliable and can leave worktrees dirty or rebase onto the wrong base.

hephaestus/automation/git_utils.py
```diff
@@ -535,6 +535,23 @@ def _remove_untracked_files_tracked_by_ref(cwd: Path, ref: str) -> list[Path]:
     return removed
 
 
+def _commit_policy_rebase_command(base_ref: str) -> list[str]:
+    """Return a rebase command that repairs signature and DCO metadata per commit."""
+    return ["git", "rebase", base_ref, "--exec", COMMIT_POLICY_REWRITE_EXEC]
```
mvillmow: This command does not guarantee that metadata is repaired. When a branch is already based on `base_ref`, `git rebase <base_ref> --exec ...` can fast-forward/no-op and skip the amend command, leaving existing branch commits without `Signed-off-by` trailers. Use `--force-rebase`/`--no-ff` or another per-commit rewrite path so the helper actually rewrites every commit it claims to enforce.

---

Re-opening: prior review comment not addressed — `_commit_policy_rebase_command()` still returns `git rebase <base_ref> --exec ...` without `--force-rebase`, `--no-ff`, or another guaranteed rewrite path.

> This command does not guarantee that metadata is repaired.

hephaestus/automation/ci_fix_orchestrator.py
```diff
@@ -403,6 +404,15 @@ def push_ci_fix(
                 return False
             if not self._ci_fix_head_is_pushable(worktree_path, issue_number):
                 return False
+        try:
+            ensure_branch_commit_metadata(worktree_path)
```
mvillmow: This uses `ensure_branch_commit_metadata()` with its default `base_branch="main"`, so a CI fix for a PR targeting another base branch is rebased onto the wrong branch before push. The existing mechanical rebase path already reads `baseRefName`; thread that same base branch into this call or resolve it from the PR before rewriting history.

---

Re-opening: prior review comment not addressed — `_push_ci_fix()` still calls `ensure_branch_commit_metadata(worktree_path)` without passing the PR base branch, so non-main PRs would still be rewritten against `origin/main`.

> This uses `ensure_branch_commit_metadata()` with its default `base_branch="main"`.

---

This rewrites HEAD after `_ci_fix_head_is_pushable()` has already validated the old clean/ahead state. A successful metadata rebase can skip/drop commits or otherwise leave HEAD no longer safely ahead of `origin/main`, and the next `push_current_branch_with_lease_on_divergence()` can still force-with-lease that rewritten HEAD to the PR branch. Re-run the pushability guard after metadata enforcement, or make the helper return a verified pushable state, before pushing.

hephaestus/automation/git_utils.py
```diff
@@ -535,6 +535,36 @@ def _remove_untracked_files_tracked_by_ref(cwd: Path, ref: str) -> list[Path]:
...
+    cwd: Path,
+    base_branch: str = "main",
+    *,
+    remote: str = "origin",
+) -> None:
+    """Rewrite branch commits so each carries a verified signature and DCO trailer."""
+    base_ref = f"{remote}/{base_branch}"
+    run(["git", "fetch", remote, base_branch], cwd=cwd)
+    try:
+        run(_commit_policy_rebase_command(base_ref), cwd=cwd)
```
mvillmow, line 563: `ensure_branch_commit_metadata()` now force-rebases after an agent turn, but it skips the stale-untracked cleanup that `rebase_worktree_onto()` runs before the same kind of rebase. `_ci_fix_head_is_pushable()` intentionally ignores untracked files, so a harmless leftover file whose path exists on the base branch can pass the pushability checks and then make this metadata rewrite fail, dropping an otherwise pushable CI fix. Call `_remove_untracked_files_tracked_by_ref(cwd, base_ref)` after the fetch and add a regression test for this helper.

mvillmow (commented):
NOGO: metadata rewrite can fail on harmless stale untracked files because it omits the existing rebase cleanup.

.github/workflows/_required.yml
```diff
@@ -420,7 +420,7 @@ jobs:
             # shellcheck disable=SC2086  # word-splitting of $bad is intentional
             printf '  %s\n' $bad
             msg="::error::Every commit MUST be cryptographically signed."
-            msg+=" Re-sign with: git rebase --exec 'git commit --amend --no-edit -S' origin/main"
+            msg+=" Re-sign with: git rebase --exec 'git commit --amend --no-edit -S -s' origin/main"
```
mvillmow, line 423: This remediation command still omits `--force-rebase`. The PR adds `_commit_policy_rebase_command()` and a regression test because plain `git rebase --exec ... origin/main` can no-op when the branch is already based on the target, so users can follow this CI hint and leave the unsigned / missing-DCO commits unchanged. Make this hint match the actual repair command, e.g. `git rebase --force-rebase origin/main --exec 'git commit --amend --no-edit -S -s'`.

scripts/check_dco_signoff.py
```diff
@@ -104,8 +104,8 @@ def main(argv: list[str] | None = None) -> int:
     if failed:
         print(
             "Every commit MUST carry a 'Signed-off-by: Name <email>' trailer "
-            "(the DCO). Add it with: git commit -s  (re-sign existing commits "
-            "with: git rebase --exec 'git commit --amend --no-edit -s' origin/main). "
+            "(the DCO). Add it with: git commit -S -s  (re-sign existing commits "
+            "with: git rebase --exec 'git commit --amend --no-edit -S -s' origin/main). "
```
mvillmow, line 108: Same remediation bug as the workflow hint: this command needs `--force-rebase` or it may not replay already-based branch commits and therefore may not run the `--exec` amend that adds the signature and DCO trailer. Update the message and the existing tests to assert the force-rebase form.

mvillmow (commented):
NOGO: DCO remediation hints can no-op without --force-rebase, contradicting the new metadata repair helper.

tests/unit/automation/test_loop_repo_manager.py
```diff
@@ -251,6 +251,48 @@ def test_returns_zero_on_empty_stdout(self, tmp_path: Path) -> None:
...
+        ):
+            sha, fetch_ok = loop_repo_manager._rebase_main("MyRepo", tmp_path)
+
+        assert (sha, fetch_ok) == ("abc1234", True)
+        assert [
+            "git",
+            "-C",
+            str(tmp_path),
+            "rebase",
+            "origin/main",
```
mvillmow, line 289: This regression test should require `_rebase_main` to use `--force-rebase` before the base ref, and the production command should be updated to match. Plain `git rebase origin/main --exec ...` can no-op when local commits are already based on `origin/main`, so `git commit --amend --no-edit -S -s` never runs and this path can still leave local commits without repaired signature/DCO metadata.

mvillmow (commented):
NOGO: loop repo metadata repair test locks in non-forced rebase, so commit sign/DCO repair can no-op.

hephaestus/automation/ci_driver.py
```diff
@@ -1323,6 +1326,16 @@ def _get_pr_branch(self, pr_number: int) -> str:
             logger.warning("Could not fetch branch for PR #%s: %s", pr_number, e)
             return f"pr-{pr_number}"
 
+    def _get_pr_base_branch(self, pr_number: int) -> str:
+        """Return the PR base branch, defaulting to ``main`` on lookup gaps."""
+        try:
+            gh_state = self._gh_pr_state(pr_number) or {}
+        except Exception as exc:
+            logger.warning("Could not fetch base branch for PR #%s: %s", pr_number, exc)
+            return "main"
```
mvillmow, line 1335: Defaulting a base-branch lookup failure to `main` is unsafe now that this value drives the pre-push metadata rebase. A transient `gh pr view` failure on a PR targeting `release/2.x` would send `push_ci_fix()` to rebase and force-push the branch against `origin/main`. Fail closed here, or only use `main` when PR data explicitly reports `baseRefName: main`.

hephaestus/automation/ci_fix_orchestrator.py
```diff
@@ -307,19 +309,32 @@ def push_ci_fix(
                 session_id=session_id,
             ):
                 return False
-        if not self._ci_fix_head_is_pushable(worktree_path, issue_number):
+        base_ref = f"origin/{pr_base_branch}"
+        if not self._ci_fix_head_is_pushable(worktree_path, issue_number, base_ref=base_ref):
```
mvillmow, line 313: This checks `origin/<pr_base_branch>` before that base ref is fetched. `sync_worktree_to_remote_branch()` fetches the PR head, while `ensure_branch_commit_metadata()` fetches `pr_base_branch` later, so a worktree without `origin/release/2.x` fails `_ci_fix_head_is_pushable()` / `_ci_fix_residual_commit_is_safe()` and aborts before reaching the fetch. Refresh the base ref before this first pushability check and add a real-path test for a missing non-main base ref.

mvillmow (commented):
NOGO: non-main base handling can rewrite against main on lookup failure or abort before fetching the base ref.

mvillmow (commented):
NOGO: broad CI-fix metadata rewrite is outside #1491, and two remediation hints omit --force-rebase.

.github/workflows/_required.yml
mvillmow, line 423: This remediation command still omits `--force-rebase`. On a branch already based on `origin/main`, `git rebase --exec ... origin/main` can no-op before running the exec; the new `_commit_policy_rebase_command` and its test show metadata repair requires forced replay. Add `--force-rebase` here.

---

The remediation hint still omits `--force-rebase`. On a branch whose commits are already based on `origin/main`, `git rebase --exec ... origin/main` may fast-forward instead of replaying each commit, so the `--exec` never runs and commits remain unsigned/un-DCO-signed. Fix: `git rebase --force-rebase --exec 'git commit --amend --no-edit -S -s' origin/main` — same form as `_commit_policy_rebase_command()`.

scripts/check_dco_signoff.py
mvillmow, line 108: This DCO remediation hint has the same no-op gap: it tells contributors to run `git rebase --exec ... origin/main` without `--force-rebase`, so existing commits may not be replayed/amended. Use the same forced-replay form as `_commit_policy_rebase_command`.

hephaestus/automation/ci_driver.py
```diff
@@ -1323,6 +1326,16 @@ def _get_pr_branch(self, pr_number: int) -> str:
...
+        base_branch = str(gh_state.get("baseRefName") or "main")
+        return base_branch or "main"
```
mvillmow, line 1337: The second `or "main"` is unreachable dead code. `str(gh_state.get("baseRefName") or "main")` always returns a non-empty string (at minimum `"main"`), so `return base_branch or "main"` can never evaluate to `"main"` via that branch. Remove the second guard: `return str(gh_state.get("baseRefName") or "main")`.

tests/unit/automation/test_loop_repo_manager.py
```diff
@@ -251,6 +251,48 @@ def test_returns_zero_on_empty_stdout(self, tmp_path: Path) -> None:
...
+                m.stdout = "origin/main\n"
+            elif "rev-list" in cmd:
+                m.stdout = "2\n"
+            elif "rev-parse" in cmd:
+                m.stdout = "abc1234\n"
+            return m
+
+        fetch_result = MagicMock(returncode=0, stderr="")
+        with (
+            patch(
```
mvillmow, line 276: This test asserts the CURRENT `_rebase_main` rebase command, which does NOT include `--force-rebase`. The PR's stated goal is that all metadata-repair rebase calls use `--force-rebase` to prevent no-op on already-based-on-target branches (see `_commit_policy_rebase_command`). `loop_repo_manager._rebase_main` is a second rebase-with-exec path that was not updated. If this omission is intentional (different semantics: `_rebase_main` rebases the main-branch checkout, not a PR branch), document why `--force-rebase` is not needed here; otherwise update `_rebase_main` and correct this assertion to match.

mvillmow (commented):
Core fix correct and well-tested. Two minor gaps: (1) unreachable double-default in _get_pr_base_branch; (2) loop_repo_manager._rebase_main still lacks --force-rebase while the new test only documents the gap.

hephaestus/automation/ci_fix_orchestrator.py
```diff
@@ -550,12 +566,8 @@ def run_ci_fix_session(
...
             ci_logs: Combined CI failure log text.
             session_id: Optional agent session ID to resume.
-            advise_findings: Prior learnings from the advise step, prepended to
-                the prompt as context. Empty or a skip marker contributes nothing.
-            pr_head_branch: The PR's head-branch name on the remote. The push
-                uses this as the destination refspec so the fix lands on the
-                actual PR branch even if the agent switched branches locally
-                during the session (#832).
+            advise_findings: Prior learnings prepended to the prompt.
+            pr_head_branch: Remote PR head branch used as the push destination.
```
mvillmow, line 573: The replacement docstring drops the `#832` cross-reference that explained *why* `pr_head_branch` is threaded separately rather than using the current local branch. That motivation (an agent can switch branches mid-session, so the push must use the original remote refspec) is non-obvious to a future reader. Suggest keeping at least: `pr_head_branch: Remote PR head branch used as the push destination, regardless of any branch the agent may have switched to locally (#832).`

mvillmow (commented):
Correct and complete: --force-rebase enforcement, pr_base_branch threading, and proportional test coverage. One minor docstring elision loses the #832 motivation for pr_head_branch separation.

hephaestus/automation/ci_fix_orchestrator.py
mvillmow: Fixed in 307a379: restored the #832 rationale on pr_head_branch — threaded explicitly because agents can switch branches mid-session; the push must target the PR's original remote refspec.

mvillmow:
**Operator GO decision** (#1554-churn pattern): reviewer verdict GO on substance in R1/R2 with threads decreasing 2→1; the final thread (restore the #832 docstring rationale) is fixed in 307a379 and resolved. Strict full-coverage audit: GO (B+, 87%) with zero blocking findings. Arming squash auto-merge per policy.

Issue #1623, opened 2026-06-26
[automation] Loop parallelizes same-file refactors → mutual merge conflicts strand PRs


mvillmow:
## Summary

Five `state:plan-go` refactor issues (#1383, #1384, #1381, #1385, #1392) were all
implemented by the automation loop in parallel, but four of their PRs went
`CONFLICTING/DIRTY` and one failed CI on a real bug. None merged; all were
hand-braked with `state:skip` and permanently stranded until manual intervention.

This issue documents the failure and proposes a fix so it does not recur.

## What happened

All five refactors targeted the **same shared file region** of
`hephaestus/automation/`:

| Issue / PR | Non-test files touched |
|---|---|
| #1383 / #1613 | `_review_utils.py`, **`address_review.py`**, `ci_driver.py` |
| #1384 / #1615 | `_review_phase.py`, **`address_review.py`**, `git_utils.py`, `implementer_phase_runner.py` |
| #1381 / #1612 | `_review_utils.py`, **`address_review.py`**, `ci_driver.py`, `plan_reviewer.py`, `pr_reviewer.py` |
| #1385 / #1616 | `_review_utils.py`, **`address_review.py`**, `ci_driver.py`, `pr_reviewer.py` |
| #1392 / #1621 | `_review_utils.py`, **`address_review.py`**, `ci_driver.py`, +6 more |

**Every PR edits `address_review.py`; four of five edit both `_review_utils.py` and
`ci_driver.py`.** They cannot land in parallel — the first to merge wins, and the
rest immediately go `CONFLICTING/DIRTY`.

Final observed state:
- #1613, #1612, #1616, #1621 → `CONFLICTING/DIRTY` (mutual merge conflicts).
- #1615 → `MERGEABLE/BLOCKED` on a **real mypy bug** (separate from the conflicts):
  `tests/unit/automation/test_address_review.py:311` does
  `assert reviewer._commit_if_changes(...) is None`, which mypy rejects with
  `error: "_commit_if_changes" ... does not return a value [func-returns-value]`.
  That cascaded: `lint` FAILURE → `required-checks-gate` FAILURE → `pr-policy` FAILURE.

When the loop's conflict recovery gave up, the operator manually applied
`state:skip` to stop the loop churning on the dead PRs. `state:skip` then also
prevents the loop from ever re-selecting those issues — so they were stranded.

## Why

- **No file-overlap awareness in issue selection.** The loop dispatches issues
[...]

PR #1776, opened 2026-07-02
fix(automation): handle final-round overlap deferrals

### LINKED ISSUE #1762: Follow-up from #1623: 1 item(s) (safety)
_Consolidated follow-up from implementation of #1623._

### Final-round file-overlap deferrals silently stranded in run_loop

loop_runner.py:998-1007 defers file-overlapping issues expecting the next --loops round to redispatch them, and RepoResult.produced_work (loop_runner.py:293) keeps the loop alive to do so. But that keep-alive only fires while loop_idx < cfg.loops (early-exit guard at loop_runner.py:1516-1517). When the deferral happens on the FINAL iteration (e.g. --loops 1, or the last loop of any run), there is no next round: the deferred issues are dropped with only an INFO log (loop_runner.py:1001-1006) and never dispatched. Repro: run with --loops 1 and two state:plan-go issues whose plans touch the same hephaestus/automation/ file — the second is deferred and then silently lost, defeating #1623's guarantee that the peer eventually runs against merged trunk. Sketch fix: in run_loop, after the loop completes, detect any RepoResult on the terminal round with non-empty deferred_issues and either (a) run one extra forced round for just those issues, or (b) surface them as a WARNING-level runner outcome so the operator knows work was withheld and must re-run. Add a run_loop-level test asserting a final-round deferral is not silently dropped.

---

mvillmow:
## Summary

Implements issue #1762 **option (b)** for final-round file-overlap deferrals: the terminal-round catch-up now runs **after** `_run_post_loop_stages` (the drive-green sweep in which the overlapping peers' PRs actually merge) and keeps the #1623 overlap protection **on** — it never redispatches an issue whose overlapping peer still owns an open PR.

### Design (per issue #1762, option (b))

- `run_loop` replays terminal-round `deferred_issues` only after the post-loop drive-green sweep, via `_run_terminal_deferred_issues`.
- The overlap guard is never forced off: catch-up dispatches use `serialize_file_overlap` as configured (default ON), `issues=[N]`, `max_workers=1`.
- Before each dispatch, `_open_peer_pr_claims` checks whether the deferred issue's planned files still overlap an open peer issue that owns an **open** PR (`find_pr_for_issue` searches open PRs only). If so, the issue is **not** redispatched — redispatching would recreate the exact same-file conflict race the deferral exists to prevent — and a prominent WARN in the loop summary names the blocking peer ("peer #N PR still open"). Nothing is silently dropped.
- The rate budget (`_maybe_sleep_for_rate_budget`) is consulted between serial catch-up dispatches.
- Catch-up `RepoResult`s are tagged `is_catchup=True` (mirroring `is_post_loop`) and excluded from per-loop convergence counting and `loops_run` accounting in both `run_loop` and `main`.

### Tests (`tests/unit/automation/test_loop_runner_file_overlap.py`)

- Reviewer's NOGO scenario: an issue whose overlapping peer PR is still open is NOT redispatched; it stays deferred with a WARN naming the peer.
- Catch-up runs strictly after `_run_post_loop_stages` (call-order assertion) with the guard ON and results tagged `is_catchup`.
- A peer whose PR already merged releases its file claim (catch-up proceeds).
- Shutdown mid-catch-up stops cleanly before the next dispatch.
- `_open_peer_pr_claims` unit tests: only open-PR peers with plans claim files; listing failures fail open.

Closes #1762


hephaestus/automation/loop_runner.py
```diff
@@ -1456,6 +1456,64 @@ def _run_post_loop_stages(cfg: LoopConfig, repos: list[str]) -> list[RepoResult]
...
+    File-overlap serialization normally relies on the next configured loop to
+    pick up deferred siblings. On the terminal loop there is no next iteration,
+    so replay the withheld issue numbers one at a time through the normal
+    repo processor. Serial replay disables the overlap guard for the forced
+    catch-up and lets each issue re-enter via the same filtering/rebase path as
+    an ordinary loop body.
+    """
+    deferred_by_repo: dict[str, list[int]] = {}
+    for result in terminal_results:
+        if result.deferred_issues:
```
mvillmow, line 1543: This collects terminal deferrals even when the same RepoResult has failures. In the normal non-terminal path, a failed overlapping peer remains open and can keep the deferred issue serialized again next loop; this forced catch-up later runs the deferred issue alone with serialize_file_overlap disabled, so it can start before the conflicting peer has merged and reintroduce the #1623 conflict class. Only replay deferrals from successful terminal results, or fall back to a warning/re-run instruction when result.any_failure is true.

mvillmow (commented):
NOGO: terminal catch-up can bypass overlap serialization after failures.

hephaestus/automation/loop_runner.py
```diff
@@ -1456,6 +1456,64 @@ def _run_post_loop_stages(cfg: LoopConfig, repos: list[str]) -> list[RepoResult]
     return results
 
 
+def _run_terminal_deferred_issues(
+    cfg: LoopConfig,
+    terminal_results: list[RepoResult],
+) -> list[RepoResult]:
+    """Redispatch final-round file-overlap deferrals serially.
+
+    File-overlap serialization normally relies on the next configured loop to
```
mvillmow: This helper adds terminal catch-up replay, but the dispatch-site comment near the overlap deferral still says deferred issues re-run in the next `--loops` iteration. Update that comment to mention terminal catch-up so the documented contract matches the new behavior.

mvillmow (commented):
Acceptable; one minor doc-comment update recommended.

hephaestus/automation/loop_runner.py
```diff
@@ -1456,6 +1456,64 @@ def _run_post_loop_stages(cfg: LoopConfig, repos: list[str]) -> list[RepoResult]
...
+    pick up deferred siblings. On the terminal loop there is no next iteration,
+    so replay the withheld issue numbers one at a time through the normal
+    repo processor. Serial replay disables the overlap guard for the forced
+    catch-up and lets each issue re-enter via the same filtering/rebase path as
+    an ordinary loop body.
+    """
+    deferred_by_repo: dict[str, list[int]] = {}
+    for result in terminal_results:
+        if result.deferred_issues:
+            deferred_by_repo.setdefault(result.repo, []).extend(result.deferred_issues)
```
mvillmow, line 1544: This treats `not result.any_failure` as proof that the overlapping peer has merged, but that is not true when `drive-green` is not selected (`--phases plan,implement` is supported, and bare `LoopConfig()` defaults to plan+implement). In that mode the first issue may have only opened a PR, and this serial catch-up immediately runs the deferred overlapping issue with overlap serialization disabled, recreating the #1623 conflict class. Gate auto-redispatch on a proven successful blocking merge/drive-green result, or fall back to warning-only when the phase selection cannot prove the peer merged; add a regression for `phases=("plan", "implement")`.

mvillmow (commented):
NOGO: terminal catch-up can redispatch overlapping work before the first PR merges when drive-green is disabled.

hephaestus/automation/loop_runner.py
```diff
@@ -1466,6 +1524,7 @@ def run_loop(cfg: LoopConfig, repos: list[str]) -> list[RepoResult]:
     safety net for the rare case where the work submission itself dies.
```
mvillmow, line 1611: This disables file-overlap serialization for the catch-up run even though the safety proof is only repo-level. `RepoResult.phases` aggregates all issues and `PhaseResult` does not identify which issue merged; a different issue's successful or no-op `drive-green` can make `_terminal_result_proves_blocking_merge()` return true while the peer that overlaps the deferred issue has not merged. That reintroduces the #1623 conflict class. Keep serialization enabled and warn, or track issue-level merge proof for the specific blocking peer before forcing `serialize_file_overlap=False`.

mvillmow (commented):
NOGO: catch-up can bypass overlap serialization without issue-level merge proof; parser change is unrelated scope creep.

hephaestus/automation/loop_runner.py
```diff
@@ -1461,6 +1466,141 @@ def _run_post_loop_stages(cfg: LoopConfig, repos: list[str]) -> list[RepoResult]
...
+                    blocker,
+                )
+                blocked.append(f"{repo}#{issue} (peer #{blocker} PR still open)")
+                continue
+
+            if catchup_results:
+                # (0, 1) keeps the rate guard active between serial dispatches:
+                # its ``loop_idx >= total_loops`` short-circuit is meant for the
+                # terminal loop boundary, and more work still follows here.
+                _maybe_sleep_for_rate_budget(0, 1)
```
mvillmow, line 1589: Passing `(0, 1)` abuses the `_maybe_sleep_for_rate_budget` API to force the guard to always run (`0 >= 1` is always False). The comment explains the intent, but consider a named sentinel or helper (`_rate_guard_between_dispatches()`) so the next reader doesn't need to reverse-engineer the arithmetic to understand why these specific magic values were chosen.

mvillmow (commented):
Correctly implements terminal deferred-issue catch-up with overlap guard kept ON, per-peer PR checks, and comprehensive tests; one minor POLA nit on the rate-guard magic constant.

mvillmow (commented):
Correct and complete option-(a) catch-up implementation. All new branches are unit-tested. One minor comment ambiguity on _ENFORCE_RATE_GUARD.

PR #1792, opened 2026-07-04
fix(docs): publish complete API reference

### LINKED ISSUE #1708: Follow-up from #1453: 1 item(s) (core)
_Consolidated follow-up from implementation of #1453._

### deploy-docs publishes near-empty pdoc reference (lazy imports skip subpackages)

pixi.toml:21 defines `docs = "pdoc hephaestus --output-dir docs/api"`. Because hephaestus/__init__.py uses PEP 562 lazy imports (`__getattr__` + `_LAZY_IMPORTS`, lines 34-60), the 19 subpackages are not real attributes at import time, so pdoc documents ONLY the top-level module. Verified: `pixi run docs` emits 3 files (index.html, search.js, hephaestus.html) with zero subpackage pages; explicitly enumerating subpackages (`pdoc hephaestus hephaestus.utils hephaestus.io ... hephaestus.agents`) emits 20 HTML pages — full coverage. The deploy-docs job (release.yml:243-249) therefore publishes a near-empty 'API reference' to GitHub Pages, defeating issue #1453. Fix: change the `docs` task to enumerate subpackages (or glob the package dir for submodules) so pdoc recurses the full tree; keep hephaestus.automation excluded if its import surface breaks the walk. Add a guard asserting docs/api contains >1 subpackage page.

---

mvillmow:
## Summary
Generate and validate a complete pdoc API reference so release docs no longer publish only the top-level lazy-imported package page.

## Changes
- Updated the docs task to enumerate the package API surface for pdoc instead of relying on lazy imports to recurse subpackages.
- Added an API reference validation helper and release workflow guard to catch near-empty generated docs.
- Adjusted related documentation and compatibility references alongside cleanup of affected automation and NATS configuration tests.

## Testing
- Added unit coverage in tests/unit/validation/test_api_reference_docs.py and updated affected automation and NATS unit tests.

## Closes
Closes #1708

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -35,8 +35,9 @@ Auto-generated API documentation is published to GitHub Pages on every release:
 - **Browse online:** <https://homericintelligence.github.io/ProjectHephaestus/>
 
 The published reference covers full function signatures, docstrings, and type
-annotations for all public modules and 52 CLI entry points, regenerated from
-the released package via [pdoc](https://pdoc.dev/).
+annotations for documented first-level subpackages, excluding the
+`hephaestus.automation` product layer, plus 52 CLI entry points, regenerated
+from the released package via [pdoc](https://pdoc.dev/).
 
 To build the same reference locally (output to the git-ignored `docs/api/`):
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -35,7 +35,7 @@ Auto-generated API documentation is published to GitHub Pages on every release:
 - **Browse online:** <https://homericintelligence.github.io/ProjectHephaestus/>
 
 The published reference covers full function signatures, docstrings, and type
-annotations for all public modules and 51 CLI entry points, regenerated from
+annotations for all public modules and 52 CLI entry points, regenerated from
 the released package via [pdoc](https://pdoc.dev/).
 
 To build the same reference locally (output to the git-ignored `docs/api/`):
--- a/hephaestus/validation/docs/api_reference.py
+++ b/hephaestus/validation/docs/api_reference.py
@@ -1,0 +1,157 @@
+"""Validate generated pdoc API reference output.
+
+The release workflow publishes ``docs/api`` to GitHub Pages after running the
+``docs`` Pixi task. A lazy top-level ``hephaestus`` import can generate only
+``hephaestus.html`` unless pdoc is pointed at the subpackages explicitly, so
+this guard verifies the generated tree contains real subpackage pages before
+upload.
+
+Usage::
+
+    hephaestus-check-api-reference
+    hephaestus-check-api-reference --docs-dir docs/api
+    hephaestus-check-api-reference --json
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from dataclasses import asdict, dataclass
+from pathlib import Path
+
+from hephaestus.cli.utils import create_validation_parser, resolve_repo_root
+
+DEFAULT_MIN_SUBPACKAGE_PAGES = 2
+PACKAGE_NAME = "hephaestus"
+
+
+@dataclass(frozen=True)
+class ApiReferenceFinding:
+    """A single generated API-reference violation."""
+
+    # kind values: "missing-docs-dir" | "too-few-subpackage-pages"
+    kind: str
+    detail: str
+
+
+def expected_pdoc_targets(repo_root: Path, package_name: str = PACKAGE_NAME) -> tuple[str, ...]:
+    """Return the pdoc targets needed to cover the package's direct subpackages.
+
+    Args:
+        repo_root: Repository root containing the package directory.
+        package_name: Top-level Python package to document.
+
+    Returns:
+        Tuple beginning with ``./<package_name>`` followed by every direct
+        subpackage as ``./<package_name>/<subpackage>`` in sorted order. Path
+        targets force pdoc to inspect the checked-out source tree even when the
+        active environment has a different editable install for the same module
+        name.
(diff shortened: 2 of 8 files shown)


hephaestus/validation/docs/api_reference.py
```diff
@@ -0,0 +1,157 @@
...
+    if not docs_dir.is_dir():
+        return [
+            ApiReferenceFinding(
+                kind="missing-docs-dir",
+                detail=f"{docs_dir} does not exist",
+            )
+        ]
+
+    subpackage_pages = list_subpackage_pages(docs_dir)
+    if len(subpackage_pages) < min_subpackage_pages:
```
mvillmow: `find_violations()` only checks that at least two direct subpackage pages exist. That passes a docs tree containing `utils.html` and `io.html` while missing the rest of the pdoc targets, so the release workflow can still upload an incomplete API reference. Build the expected direct subpackage page set from `expected_pdoc_targets()` or a shared discovery helper, then emit a `missing-subpackage-page` finding for each absent page.

mvillmow (commented):
NOGO: API reference guard only enforces two pages, not full subpackage coverage.

pixi.toml
```diff
@@ -18,7 +18,7 @@ mypy-pkg = "mypy hephaestus/"
 # Runs pip-audit in the lint environment without a nested `pixi run` invocation.
 # The top-level `audit` alias was removed: use `pixi run --environment lint audit`
 # or `pixi run --environment lint pip-audit` directly (as CI already does).
-docs = "pdoc hephaestus --output-dir docs/api"
+docs = "pdoc ./hephaestus ./hephaestus/agents ./hephaestus/automation ./hephaestus/benchmarks ./hephaestus/ci ./hephaestus/cli ./hephaestus/config ./hephaestus/datasets ./hephaestus/discovery ./hephaestus/forensics ./hephaestus/github ./hephaestus/io ./hephaestus/logging ./hephaestus/markdown ./hephaestus/nats ./hephaestus/resilience ./hephaestus/scripts_lib ./hephaestus/system ./hephaestus/utils ./hephaestus/validation ./hephaestus/version --output-dir docs/api"
```
mvillmow: This includes `./hephaestus/automation` in the generated API reference, but issue #1708 scopes the fix to documented first-level subpackages while keeping `hephaestus.automation` excluded. Because the new validator also derives expected pages from every direct subpackage, this PR now requires `automation.html` instead of preserving the exclusion. Remove automation from the pdoc target set, apply the same exclusion in the helper that derives expected pages, and add a test that locks that exclusion.

---

This docs task now includes `./hephaestus/automation`, but #1708 scoped the generated API reference to the documented library subpackages with the automation product layer excluded. `hephaestus.automation` is the opt-in product layer per ADR-0001, and pdoc introspection of that package emits unresolved `__all__` warnings. Remove this target and make the target-discovery helper exclude `automation` as well.

mvillmow (commented):
NOGO: docs generation/guard includes automation despite the issue scope requiring it to stay excluded.

hephaestus/validation/docs/api_reference.py
mvillmow: This guard only checks that at least two direct subpackage pages exist, so a broken `docs/api` containing only `hephaestus/utils.html` and `hephaestus/io.html` passes even if most expected subpackage pages are missing. The issue is about publishing the full first-level API reference; build the expected page set from the same discovered pdoc targets, excluding `automation`, and report each missing page instead of accepting any count above two.

mvillmow (commented):
NOGO: docs task includes automation and guard accepts incomplete subpackage coverage.

docs/index.md
```diff
@@ -35,7 +35,7 @@ Auto-generated API documentation is published to GitHub Pages on every release:
 - **Browse online:** <https://homericintelligence.github.io/ProjectHephaestus/>
 
 The published reference covers full function signatures, docstrings, and type
-annotations for all public modules and 51 CLI entry points, regenerated from
+annotations for all public modules and 52 CLI entry points, regenerated from
```
mvillmow: The generated pdoc target list intentionally excludes `hephaestus.automation`, but this sentence still says the published reference covers "all public modules" while this page also lists automation as a subpackage. Update the wording to say the API reference covers documented first-level subpackages except the automation product layer.

hephaestus/validation/docs/api_reference.py
```diff
@@ -0,0 +1,172 @@
...
+    """Entry point for ``hephaestus-check-api-reference``."""
+    parser = create_validation_parser(__doc__, prog="hephaestus-check-api-reference")
+    parser.add_argument(
+        "--docs-dir",
+        type=Path,
+        default=None,
+        help="Generated API reference directory (default: <repo-root>/docs/api)",
+    )
+    parser.add_argument(
+        "--min-subpackage-pages",
```
mvillmow: This newly introduced `--min-subpackage-pages` flag is parsed but ignored, so callers can pass a value that has no effect. Remove the argument, or make it actually control validation behavior; if it is only retained for compatibility, hide it from help and document the existing consumer.

---

This flag is accepted but never read after parsing, so `--min-subpackage-pages 999` still runs the same full-page validation. Remove the unused argument, or wire it into validation, so the CLI does not expose a silent no-op option.

mvillmow (commented):
GO: fixes docs generation and release guard; only minor cleanup around wording and an unused flag.

mvillmow (commented):
LGTM

PR #1803, opened 2026-07-04
[Docs] Bump MIGRATION.md latest-version to 0.9.9 ahead of the release tag

### LINKED ISSUE #1802: Bump MIGRATION.md latest-version line ahead of next release tag (v0.9.9)
## Problem

The v0.9.8 release failed in the Release workflow's test job: `test_migration_md_version_does_not_trail_latest_git_tag` requires `docs/MIGRATION.md`'s "latest released version is **X.Y.Z**" line to be >= the latest git tag, but the v0.9.8 tag's tree still documents 0.9.7. Tags are immutable, so v0.9.8 cannot be released; the recovery is to bump the doc and cut v0.9.9 (v0.9.8 remains an unpublished tag).

This is the same trap that broke v0.9.7's first release attempt (fixed by #1542). The established release order is: bump the doc, merge, then tag — but `docs/RELEASING.md`'s pre-release checklist doesn't mention it and even claims "The version itself does not need to be edited in any file."

## Fix

1. Bump `docs/MIGRATION.md` line 3 to "latest released version is **0.9.9**" (and refresh the as-of date).
2. Add the doc-bump step to `docs/RELEASING.md`'s Pre-Release Checklist so the next release doesn't hit it. (Longer term this belongs in the auto-tag workflow itself — see #1801 for the related broken auto-tag → release trigger chain.)

## Acceptance

- Release workflow test job passes on the next tag.

mvillmow:
## Summary

Bumps `docs/MIGRATION.md`'s "latest released version" line to **0.9.9** so the Release workflow's `test_migration_md_version_does_not_trail_latest_git_tag` guard passes on the upcoming v0.9.9 tag. The v0.9.8 tag is stranded unreleasable: its (immutable) tree still documents 0.9.7, which is exactly how this guard failed the v0.9.8 release run (28693531484).

Also adds the bump-doc-then-tag step to the RELEASING.md Pre-Release Checklist so the trap is documented instead of rediscovered each release (it previously broke the first v0.9.7 attempt too; cf. #1542).

## Test plan

- `pixi run pytest tests/unit/docs/test_version_currency.py` passes with the doc claiming 0.9.9 (guard is `documented >= latest tag`; 0.9.9 > 0.9.8 tag).
- Markdown lint via pre-commit passed.

Closes #1802

### DIFF
--- a/docs/MIGRATION.md
+++ b/docs/MIGRATION.md
@@ -1,6 +1,6 @@
 # Migration Guide
 
-> **Status (as of 2026-06-20):** The latest released version is **0.9.7** (tag-driven
+> **Status (as of 2026-07-03):** The latest released version is **0.9.9** (tag-driven
 > via hatch-vcs). **1.0 has not been released yet** — the section below is the
 > *forthcoming* 1.0 migration guidance, published ahead of the cut so consumers can
 > prepare. There are currently **no breaking changes between 0.9.x releases**: upgrading
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -32,8 +32,13 @@ Before triggering the workflow, ensure:
 - [ ] All changes merged to `main` and CI is green.
 - [ ] `pixi.lock` is up to date (`pixi install` produces no changes).
 - [ ] No open issues in the milestone you are releasing.
+- [ ] `docs/MIGRATION.md`'s "latest released version is **X.Y.Z**" line already names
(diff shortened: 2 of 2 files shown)


mvillmow (commented):
LGTM — MIGRATION.md bumped to 0.9.9 with refreshed date, RELEASING.md checklist updated with accurate warning. Minimal, targeted, no issues.

PR #1825, opened 2026-07-04
fix(automation): log claude stderr on resume failures

### LINKED ISSUE #1799: advise/claude resume failures swallow stderr — real failure reason never logged
## Problem

`hephaestus/automation/advise_runner.py` catches all exceptions from the advise claude call (lines 516-518):

```python
except Exception as e:
    logger.warning("Advise step failed for issue #%s: %s", issue_number, e)
```

For a `subprocess.CalledProcessError`, `str(e)` is only "Command '[...]' returned non-zero exit status N." — `e.stderr` (the actual reason, e.g. a resume "No conversation found" for the deterministic uuid5 session id built in `claude_invoke.py:127-139`) is never included. In the 2026-07-03 run (output2.log:282) the advise step for issue #121 failed in ~3s with mode=resume and the log carries zero diagnostic information.

## Fix

- In `run_advise`'s handler (and an audit of the other `except Exception` catches around `invoke_claude_with_session` callers): when `e` is a `CalledProcessError`, include a truncated `e.stderr` in the warning.
- Optional hardening: if the resume failure's stderr matches the session-not-found phrasings (`SESSION_EXPIRED_PHRASES` in `claude_invoke.py:40-49` — note "No conversation found" is missing from that tuple and should be added), retry once in create mode (`--session-id`). This reinstates a *narrow* recreate path only for a vanished transcript, not the removed general recreate cascade (#1168).

## Acceptance

- An advise resume failure logs the underlying stderr.
- (If hardening adopted) a deleted transcript no longer permanently skips advise for that issue.

mvillmow:
## Summary
Expose the underlying Claude CLI stderr/stdout when resume-mode automation calls fail, so advise, implementer, and post-merge runners log the real diagnostic reason instead of a bare `CalledProcessError` exit-code string.

## Changes
- `hephaestus/automation/claude_invoke.py`: add `format_called_process_error` — a bounded (500-char truncated), bytes-safe stderr/stdout formatter — and `describe_claude_failure(exc)`, a shared helper that dispatches `CalledProcessError` to the detailed formatter and everything else to `str(exc)` (DRY: replaces three copies of the same inline `isinstance` pattern).
- `hephaestus/automation/advise_runner.py`: use `describe_claude_failure` for the failure warning, and embed the detailed failure in the `advise_skipped` prompt marker too (previously a bare `str(e)`, which for `CalledProcessError` is just an exit code).
- `hephaestus/automation/implementer_phase_runner.py`: dirty-worktree decision failure log now carries stderr detail via the shared helper.
- `hephaestus/automation/post_merge_processor.py`: drive-green `/learn` failure log now carries stderr detail via the shared helper.
- Extend Claude session-expired detection coverage ("No conversation found").

## Testing
- New direct edge tests for `format_called_process_error` (bytes streams, invalid UTF-8 replacement, 500-char truncation boundary, stdout-only, no-stream fallback, custom `max_chars`) and for `describe_claude_failure` dispatch in `tests/unit/automation/test_claude_invoke.py`.
- Call-site tests assert the stderr text reaches the warning logs (advise, implementer loop, post-merge) and the advise skip marker itself.
- `pixi run pytest tests/unit/automation -k "claude_invoke or advise or post_merge or implementer_loop"`: 251 passed. `pixi run mypy`, `pixi run ruff check`, and `pre-commit run --all-files` all clean.

Closes #1799

Generated by Codex via ProjectHephaestus automation; consolidated per strict-audit review.

### DIFF
--- a/hephaestus/automation/advise_runner.py
+++ b/hephaestus/automation/advise_runner.py
@@ -44,6 +44,7 @@ except ModuleNotFoundError:  # pragma: no cover - Windows fallback
     # name to None on the Windows fallback path needs [assignment].
     fcntl = None  # type: ignore[assignment]
 
+from .claude_invoke import format_called_process_error
 from .git_utils import get_repo_root
 
 logger = logging.getLogger(__name__)
@@ -66,6 +67,13 @@ class SelectedSkill:
     source: str
     reason: str
     path: Path
+
+
+def _advise_failure_detail(exc: Exception) -> str:
+    """Return the diagnostic string for a failed advise run."""
+    if isinstance(exc, subprocess.CalledProcessError):
+        return format_called_process_error(exc)
+    return str(exc)
 
 
 def advise_skipped(reason: str) -> str:
@@ -515,5 +523,12 @@ def run_advise(
             selected = parse_selected_skills(selector_output, mnemosyne_root)
         return format_selected_skill_context(selected)
     except Exception as e:
-        logger.warning("Advise step failed for issue #%s: %s", issue_number, e)
+        detail = _advise_failure_detail(e)
+        logger.warning(
+            "Advise step failed for issue #%s: %s",
+            issue_number,
+            detail,
+            extra={"exception_type": type(e).__name__},
+            exc_info=True,
+        )
         return advise_skipped(f"unexpected error: {e}")
--- a/hephaestus/automation/advise_runner.py
+++ b/hephaestus/automation/advise_runner.py
@@ -44,7 +44,7 @@ except ModuleNotFoundError:  # pragma: no cover - Windows fallback
     # name to None on the Windows fallback path needs [assignment].
     fcntl = None  # type: ignore[assignment]
 
-from .claude_invoke import format_called_process_error
+from .claude_invoke import describe_claude_failure
 from .git_utils import get_repo_root
 
 logger = logging.getLogger(__name__)
@@ -67,13 +67,6 @@ class SelectedSkill:
     source: str
     reason: str
     path: Path
-
-
-def _advise_failure_detail(exc: Exception) -> str:
-    """Return the diagnostic string for a failed advise run."""
-    if isinstance(exc, subprocess.CalledProcessError):
-        return format_called_process_error(exc)
-    return str(exc)
 
 
 def advise_skipped(reason: str) -> str:
@@ -523,7 +516,7 @@ def run_advise(
             selected = parse_selected_skills(selector_output, mnemosyne_root)
         return format_selected_skill_context(selected)
     except Exception as e:
-        detail = _advise_failure_detail(e)
+        detail = describe_claude_failure(e)
         logger.warning(
             "Advise step failed for issue #%s: %s",
             issue_number,
@@ -531,4 +524,7 @@ def run_advise(
             extra={"exception_type": type(e).__name__},
             exc_info=True,
         )
-        return advise_skipped(f"unexpected error: {e}")
+        # The marker gets embedded in downstream prompts/plan bodies, so it
+        # must carry the stderr detail too — a bare str(e) is just an exit
+        # code for CalledProcessError (#1799).
+        return advise_skipped(f"unexpected error: {detail}")
--- a/hephaestus/automation/claude_invoke.py
+++ b/hephaestus/automation/claude_invoke.py
@@ -44,10 +44,45 @@ from hephaestus.utils.helpers import strip_null_bytes
 
 logger = logging.getLogger(__name__)
 
+_MAX_SUBPROCESS_FAILURE_DETAIL_CHARS = 500
+
+
+def _subprocess_stream_text(value: object) -> str:
+    """Return subprocess stream content as text for diagnostic logging."""
+    if value is None:
+        return ""
+    if isinstance(value, bytes):
+        return value.decode(errors="replace")
+    return str(value)
+
+
+def _truncate_subprocess_detail(text: str, max_chars: int) -> str:
+    """Bound subprocess diagnostics before writing them to shared logs."""
+    if len(text) <= max_chars:
+        return text
+    return text[:max_chars] + "... [truncated]"
+
+
+def format_called_process_error(
+    exc: subprocess.CalledProcessError,
+    *,
+    max_chars: int = _MAX_SUBPROCESS_FAILURE_DETAIL_CHARS,
+) -> str:
+    """Format a ``CalledProcessError`` with bounded stdout/stderr diagnostics."""
+    parts = [str(exc)]
+    stderr = _subprocess_stream_text(exc.stderr).strip()
+    if stderr:
+        parts.append(f"stderr={_truncate_subprocess_detail(stderr, max_chars)!r}")
+    stdout = _subprocess_stream_text(exc.stdout).strip()
+    if stdout:
+        parts.append(f"stdout={_truncate_subprocess_detail(stdout, max_chars)!r}")
+    return "; ".join(parts)
+
 
 # Substrings the Claude CLI returns when ``--resume`` targets a session that
 # no longer exists in local persistence.
 SESSION_EXPIRED_PHRASES: tuple[str, ...] = (
+    "no conversation found",
     "session not found",
     "invalid session",
     "session expired",
--- a/hephaestus/automation/claude_invoke.py
+++ b/hephaestus/automation/claude_invoke.py
@@ -79,6 +79,19 @@ def format_called_process_error(
     return "; ".join(parts)
 
 
+def describe_claude_failure(exc: BaseException) -> str:
+    """Return a bounded diagnostic string for a failed ``claude`` invocation.
+
+    ``CalledProcessError`` carries the CLI's stderr/stdout, which is the only
+    place the actual failure reason lives (#1799) — surface it via
+    :func:`format_called_process_error`. Any other exception degrades to
+    ``str(exc)``.
+    """
+    if isinstance(exc, subprocess.CalledProcessError):
+        return format_called_process_error(exc)
+    return str(exc)
+
+
 # Substrings the Claude CLI returns when ``--resume`` targets a session that
 # no longer exists in local persistence.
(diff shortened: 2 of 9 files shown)


mvillmow (commented):
NOGO: stderr logging is addressed, but unrelated API-reference validation changes make the PR fail P2/YAGNI single-purpose review.

hephaestus/validation/docs/api_reference.py
```diff
@@ -0,0 +1,163 @@
+"""Validate generated pdoc API reference output.
```
mvillmow, line 1: P2/YAGNI: this new API-reference validation CLI and its related workflow/docs changes do not map to issue #1799, which is about preserving stderr from advise/Claude subprocess failures. Please remove this unrelated feature from this PR or move it to a separate issue/PR so this change remains single-purpose.

hephaestus/automation/claude_invoke.py
```diff
@@ -39,10 +39,45 @@
...
+    return text[:max_chars] + "... [truncated]"
+
+
+def format_called_process_error(
+    exc: subprocess.CalledProcessError,
+    *,
+    max_chars: int = _MAX_SUBPROCESS_FAILURE_DETAIL_CHARS,
+) -> str:
+    """Format a ``CalledProcessError`` with bounded stdout/stderr diagnostics."""
+    parts = [str(exc)]
```
mvillmow, line 72: `str(exc)` includes the full `CalledProcessError.cmd`; for many `invoke_claude_with_session` callers the prompt is passed as an argv element, so this warning can dump unbounded issue/prompt content while trying to add bounded stderr diagnostics. Build the message from safe fields instead, e.g. command executable plus return code, then append the truncated stderr/stdout excerpts.

---

`str(exc)` includes the full `CalledProcessError.cmd` list. For callers that pass the prompt as a positional argv element (the default path in `invoke_claude_with_session` with `input_via_stdin=False`), this dumps unbounded issue/prompt content into the log — defeating the bounded-diagnostics goal of this function. Replace `str(exc)` with safe fields only:
```python
exe = os.path.basename(exc.cmd[0]) if exc.cmd else "<unknown>"
parts = [f"Command '{exe}' returned non-zero exit status {exc.returncode}"]
```
This avoids leaking argv content while still conveying the essential failure info.

mvillmow (commented):
NOGO: stderr is captured, but the formatter can still log unbounded prompt/issue content via str(exc).

hephaestus/automation/advise_runner.py
```diff
@@ -514,5 +522,12 @@ def run_advise(
...
         return format_selected_skill_context(selected)
     except Exception as e:
-        logger.warning("Advise step failed for issue #%s: %s", issue_number, e)
+        detail = _advise_failure_detail(e)
+        logger.warning(
+            "Advise step failed for issue #%s: %s",
+            issue_number,
+            detail,
+            extra={"exception_type": type(e).__name__},
+            exc_info=True,
```
mvillmow, line 525: This `exc_info=True` defeats the prompt-safe formatting added above for `CalledProcessError`. Claude advise callers pass the selector prompt through `invoke_claude_with_session` with the default positional-argv prompt path, so the traceback can render `CalledProcessError.__str__` with the full `cmd` including untrusted issue text / marketplace prompt content. Keep the sanitized `format_called_process_error()` message, but avoid traceback logging for `CalledProcessError` here, e.g. `exc_info=not isinstance(e, subprocess.CalledProcessError)` or a separate sanitized traceback-free branch.

---

`exc_info=True` is unconditional. For `CalledProcessError`, the traceback's `__str__` renders the full `cmd` (which can include the selector prompt text injected as a positional argv argument). Use `exc_info=not isinstance(e, subprocess.CalledProcessError)` to suppress the traceback for `CalledProcessError` while still logging tracebacks for unexpected exception types. The sanitized `detail` string already carries the relevant diagnostic.

mvillmow (commented):
NOGO: advise warning can still log prompt-bearing CalledProcessError argv via exc_info=True.

mvillmow (commented):
NOGO: format_called_process_error leads with str(exc) (exposes full cmd/prompt in logs) and exc_info=True is unconditional for CalledProcessError — both findings carried over from prior review pass, still unresolved.

hephaestus/automation/post_merge_processor.py
```diff
@@ -184,10 +184,13 @@ def run_drive_green_learnings(self, issue_number: int, pr_number: int) -> bool:
             logger.info("Issue #%s: drive-green learnings captured", issue_number)
             return True
         except Exception as e:  # broad: external claude process; non-blocking
+            detail = describe_claude_failure(e)
             logger.warning(
                 "Issue #%s: drive-green learnings failed (non-fatal): %s",
                 issue_number,
-                e,
+                detail,
+                extra={"exception_type": type(e).__name__},
```
mvillmow, line 192: Same `exc_info=True` concern as `advise_runner.py`: for a `CalledProcessError`, the traceback exposes `cmd`. The `drive-green` invoke path also passes the prompt positionally, so apply the same guard: `exc_info=not isinstance(e, subprocess.CalledProcessError)`.

hephaestus/automation/claude_invoke.py
```diff
@@ -39,10 +39,58 @@
...
+_MAX_SUBPROCESS_FAILURE_DETAIL_CHARS = 500
+
+
+def _subprocess_stream_text(value: object) -> str:
+    """Return subprocess stream content as text for diagnostic logging."""
+    if value is None:
+        return ""
+    if isinstance(value, bytes):
+        return value.decode(errors="replace")
+    return str(value)
```
mvillmow, line 56: The plan spec'd single-line normalization (collapse `\n` and control chars to spaces) before truncation, because the formatted string is embedded in downstream prompts via `advise_skipped()`. Without it, a multi-line stderr like `No conversation found\nretry later` will appear verbatim in the HTML skip-marker injected into plan bodies. Add: `text = re.sub(r'[\s\x00-\x1f\x7f]+', ' ', text).strip()` before the truncation call in `format_called_process_error`, or inside `_subprocess_stream_text`.

mvillmow (commented):
Core acceptance met: CalledProcessError stderr surfaces at all 3 call sites with bounded formatting and exc_info guard. Minor: stream text is not newline/control-char normalized before embedding in advise_skipped() markers.

hephaestus/automation/claude_invoke.py
```diff
@@ -39,10 +39,58 @@
...
+) -> str:
+    """Format a ``CalledProcessError`` with bounded stdout/stderr diagnostics."""
+    parts = [str(exc)]
+    stderr = _subprocess_stream_text(exc.stderr).strip()
+    if stderr:
+        parts.append(f"stderr={_truncate_subprocess_detail(stderr, max_chars)!r}")
+    stdout = _subprocess_stream_text(exc.stdout).strip()
+    if stdout:
+        parts.append(f"stdout={_truncate_subprocess_detail(stdout, max_chars)!r}")
+    return "; ".join(parts)
```
mvillmow, line 79: Redundant `.strip()` call: `_subprocess_stream_text` already calls `.strip()` internally (line 55), so the result is already stripped. Same applies to line 77 for stdout. Harmless but noisy — remove the redundant `.strip()` calls after `_subprocess_stream_text(...)` in `format_called_process_error`.

tests/unit/automation/test_advise_runner.py
```diff
@@ -372,6 +373,41 @@ def test_degrades_to_skip_on_exception(self) -> None:
...
+                issue_body="b",
+                invoke=lambda _prompt: (_ for _ in ()).throw(err),
+                build_prompt=_build_prompt,
+            )
+
+        assert result.startswith("<!-- advise step skipped:")
+        assert "No conversation found with session ID abc" in caplog.text
+        # The prompt-embedded skip marker must carry the stderr detail too —
+        # a bare str(e) is just an exit code for CalledProcessError (#1799).
+        assert "No conversation found with session ID abc" in result
```
mvillmow, line 410: The test uses clean single-line stderr, so it doesn't verify that the skip marker is log-safe when stderr contains newlines or control characters (the plan named `test_called_process_error_warning_includes_log_safe_stderr` for exactly this). A failing scenario: `stderr="No conversation found\nretry later\x00"` — the normalization in `_subprocess_stream_text` would strip control chars before embedding, but this test won't catch a regression that removes that normalization. Consider adding a second assertion with embedded `\n`/`\x00` to pin the log-safety contract.

mvillmow (commented):
Required acceptance criterion met: CalledProcessError stderr is now logged and embedded in skip marker. Two minor issues: redundant .strip() after _subprocess_stream_text (already strips), and missing normalization regression test in advise test.

hephaestus/automation/claude_invoke.py
mvillmow: Refuted with code citation: _subprocess_stream_text (claude_invoke.py:45-51) does NOT strip — it only None-guards, decodes bytes, and str()s. The .strip() calls in format_called_process_error are therefore not redundant; removing them would leave trailing newlines inside the repr-quoted detail. No change made.

tests/unit/automation/test_advise_runner.py
mvillmow: Fixed in f26e14c: added the plan-named test_called_process_error_warning_includes_log_safe_stderr — multiline/NUL stderr is repr-escaped in both the warning log and the prompt-embedded skip marker (asserts escaped form in caplog, no raw NUL in the marker). 30 advise_runner tests green.

mvillmow:
**Operator GO decision** (#1554-churn pattern, third instance): reviewer verdict GO on substance in R1/R2, stranding only on per-round minor threads at the cap. Both final threads dispatched: the redundant-strip claim is refuted with a code citation (_subprocess_stream_text never strips), and the log-safety regression test is added in f26e14c. Strict audit: CONDITIONAL→fixed (zombie #1792-duplicate commit dropped, describe_claude_failure consolidation, edge tests). Arming squash auto-merge per policy.

PR #1831, opened 2026-07-04
docs(adr): ADR-0006 queue-based automation pipeline + architecture doc skeleton

mvillmow:
## Summary

First PR of epic #1809 (queue-based in-process automation pipeline). Establishes the written design contract before any pipeline code lands:

- **`docs/adr/0006-queue-based-in-process-automation-pipeline.md`** — Nygard-format ADR recording the decision (8 in-memory stage queues, single coordinator thread owning all GitHub mutations/routing, one worker pool for agent/build/git jobs, declarative per-stage failure routing, GitHub-labels-as-journal), the alternatives considered (subprocess-per-phase with queue façade, asyncio coordinator, on-disk queue/SQLite, single typed-message queue), and consequences.
- **`docs/AUTOMATION_LOOP_ARCHITECTURE.md`** — design-doc skeleton with `> Status: pre-implementation` banner: queue topology (mermaid + ASCII), the eight-stage table (states, ordered [M]/[W:A]/[W:B]/[W:G] steps, prompt-builder function per stage, verdicts, fail targets, budgets, owned labels), ROUTES table, and the seeding/reconstruction classification table.
- ADR index entry in `docs/adr/README.md`.

Docs-only change — no runtime code touched.

Closes #1810

Issue #1809, opened 2026-07-04
Epic: queue-based in-process automation pipeline (replaces issue-major subprocess loop)


mvillmow:
# Epic: queue-based in-process automation pipeline (replaces issue-major subprocess loop)

**Type:** Epic / architecture refactor. **Do not implement this issue directly** — all work happens in the serialized sub-issues below. This epic carries the full approved plan so the complete design lives on GitHub.

## Why

The `hephaestus-automation-loop` has accreted legacy code paths across several rewrites (phase-major → issue-major inversion #1560/PR #1561, review-loop overhaul, state:skip fixes). Too many incorrect steps come from legacy paths; the loops, worker pools, and phases are hard-linked. This epic remaps the loop into a queue-based, state-machine pipeline:

- **One in-memory work queue per stage**: `repo → planning → plan_review → implementation → pr_review → ci → merge_wait → finished`.
- **Each stage is a small state machine** that pulls from its queue, processes, and pushes to the next queue on success or a declared fail-target queue on failure (per-stage routing table, default = previous queue).
- **The coordinator (main thread) does ONLY**: arg parsing + queue seeding, draining queues, validation/logging/GitHub API manipulation, routing. It never launches agent workflows or build/test commands.
- **A single worker pool** runs ALL agent invocations, build/test subprocesses, and git/network operations.
- **Seeding**: selected repos → repo queue; selected issues → planning queue (with label fast-forward); selected PRs → implementation-path entry (ci or pr_review by `state:implementation-go`).
- **GitHub is the journal**: queues are in-memory and reconstructed from labels/PR state at startup; every durable write precedes the corresponding queue push. Interrupt → items are RESUMABLE, never FAILED; on interrupt or completion a results summary prints to screen; final stage threshold → finished queue marked pass/fail.

## Delivery

Feature-flagged (`--pipeline` / `HEPH_PIPELINE=1`, default OFF) across 9 strictly serialized PRs, then a cutover default-flip, then a 5-issue cleanup wave deletes the legacy path after ≥1 release of live runs. Net coverage-omit-list delta −4..−6 (issue #1422 trajectory). Every sub-issue carries `Depends on #<prev>` — the loop must never work two of these concurrently.

## Sub-issues

| # | Issue | Depends on |
|---|-------|------------|
| 1 | #1810 docs(adr): ADR-0006 queue-based in-process automation pipeline + architecture doc skeleton | — |
[...]

PR #1833, opened 2026-07-04
feat(pipeline): work items, stage queues, failure routing table

### LINKED ISSUE #1811: feat(pipeline): work items, stage queues, failure routing table
Part of epic #1809.

Prerequisite: #1810 (merged 2026-07-04).

## Context

Foundation layer of the pipeline: the core data types and the declarative routing table, as pure data/functions with **zero I/O** — no `gh`, no `claude`, no subprocess, no imports from `github_api`/`claude_invoke`. Fully unit-tested; these modules must NOT enter the coverage omit list.

## Deliverables

New package `hephaestus/automation/pipeline/` (PEP 562 lazy `__init__.py` per ADR-0002):

1. **`work_item.py`** — `ItemKind(str, Enum)` (REPO/ISSUE/PR); `HistoryEvent` (ts, stage, state, note; history capped at 200/item); `ItemResult` (passed, reason, final_stage); `WorkItem` dataclass: repo, kind, issue, pr, `stage: StageName`, `state: str` (stage-local mini-state, in-memory only — never a GitHub label), `attempts: dict[str,int]` (counters: plan, plan_review_iter, plan_cycles, pr_review_iter, pr_review_hard, test_fix, ci_fix, blocked_address, rebase, merge), history, timestamps, worktree/branch, session_ids, labels cache, payload scratch dict, result. Document the thread-safety rule in the module docstring: **a WorkItem is only ever touched by the coordinator thread**.
2. **`queues.py`** — `StageQueue`: FIFO deque owned exclusively by the coordinator, deliberately NOT thread-safe, with push/pop/len/snapshot; `CompletionQueue = queue.Queue` type alias (the only cross-thread channel, used from #1809 PR-3 onward).
3. **`routing.py`** — `StageName(str, Enum)` (repo, planning, plan_review, implementation, pr_review, ci, merge_wait, finished); `Route(next, fail_routes: dict[str, StageName] with "*" default, budgets: dict[str,int])`; the `ROUTES` table mapping today's caps (plan_review_iter 3, plan_cycles 2, pr_review_iter 3, pr_review_hard 6, ci_fix 1, blocked_address 2, rebase 2, merge = max_merge_attempts); `Disposition` enum (ADVANCE/RETRY/FAIL_BACK/SKIP/BLOCKED/FINISH_PASS/FINISH_FAIL) + `StageOutcome`; `PipelineScope` that trims ROUTES to a stage subset (last in-scope stage routes to finished; fail routes exiting the scope become finished).
4. **Bank the stale omit entry**: `hephaestus/automation/github_api.py` no longer exists on disk (flat module became the `github_api/` package) but is still in `[tool.coverage.run].omit`. Remove it from `pyproject.toml`, from `expected_modules` in `tests/unit/validation/test_omit_allowlist.py` (fix the "17 automation modules" comment → 16), and the count in the `tests/integration/test_orchestration_smoke.py` docstring.

## Tests (new: `tests/unit/automation/pipeline/`)

- `test_work_item.py`, `test_queues.py`, `test_routing.py` — table-driven.
- `test_routing_properties.py` — hypothesis (already a dev dependency): (a) every non-terminal stage has a `"*"` fail route; (b) every cycle in the routing graph passes through an attempt-budget key; (c) driving any item through budget+1 failures lands it at the fail target, never back in the loop; (d) every `PipelineScope` subset yields routes that stay inside the scope ∪ {finished}.

## Acceptance criteria

- [ ] Zero I/O imports in the new modules (assert with a simple import-graph test).
- [ ] New modules NOT in the coverage omit list; coverage over them reported > 0%.
- [ ] Omit-list triple edit done (pyproject + test_omit_allowlist + smoke docstring); `pixi run pytest tests/unit/validation tests/integration/test_orchestration_smoke.py -q` green.
- [ ] `pixi run mypy` whole-tree clean (package fully typed from day one); ruff + pre-commit clean.
- [ ] PR body `Closes #1811`.

mvillmow:
## Summary

Second PR of epic #1809 — the pure-data foundation layer of the queue-based in-process automation pipeline, implementing the contract in ADR-0006 and `docs/AUTOMATION_LOOP_ARCHITECTURE.md` (merged in #1831):

- **`hephaestus/automation/pipeline/work_item.py`** — `WorkItem` (repo/kind/issue/pr, stage, in-memory state, per-item-lifetime `attempts` counters for all 12 ROUTES budget keys, history capped at `HISTORY_CAP=200`, tz-aware UTC timestamps), `HistoryEvent`, `ItemResult`, `ItemKind`. Plain dataclasses, coordinator-thread-only by contract, no locks.
- **`hephaestus/automation/pipeline/queues.py`** — `StageQueue` (coordinator-owned deque, introspectable) and `CompletionQueue` (the single cross-thread `queue.Queue` channel; payload becomes `(JobHandle, JobResult)` when the worker-pool slice lands).
- **`hephaestus/automation/pipeline/routing.py`** — `StageName`, `Disposition`, `Route`, and a `ROUTES` table that transcribes the architecture doc's normative table **exactly**: repo item terminal with `clone=2`, `plan=2`, `plan_review_iter=3`/`plan_cycles=2`, `implement=2`/`test_fix=1`, `pr_review_iter=3`/`pr_review_hard=6`, `ci_fix=1`/`rebase=2`, `blocked_address=2`/`rebase=2`/`merge=--max-merge-attempts` — including every named fail-route reason key (`plan_not_go`, `already_implementation_go_pr`, `agent_error`, `human_blocked`, `exhaustion`, `fix_exhausted`, `not_implementation_go`, `no_pr`, `ci_red`, `blocked_exhausted`, `closed`, `timeout`, `nogo`, `plan_cycles_exhausted`). `PipelineScope` validates non-empty contiguous scopes and returns defensive copies.
- **Zero I/O** in the subpackage — enforced by a recursive AST guard that also forbids `hephaestus.utils`/`hephaestus.resilience` (subprocess wrappers) and `asyncio`/`pty`/`fcntl`/`tempfile`.
- **Tests (74)**: table tests per module, a full-table doc-conformance pin, REAL hypothesis property tests (driven failure-sequence termination over generated reason lists, per-budget exhaustion, generated contiguous-scope closure), an `attempts`↔`ROUTES`-budgets lockstep test, and mutable-default non-aliasing tests. Modules fully covered — NOT on the coverage omit list.
- Banks the stale `hephaestus/automation/github_api.py` omit entry (file no longer exists): net omit delta −1, with lockstep `test_omit_allowlist.py` and orchestration-smoke edits.

Per ADR-0001, everything stays inside the `hephaestus/automation/` product layer; the base import surface is untouched (guards green).

Strict-audit trail: wave-1 findings (ROUTES doc-drift, missing hypothesis, scope aliasing) fixed in the second commit.

Closes #1811


tests/unit/automation/pipeline/test_queues.py
```diff
@@ -0,0 +1,107 @@
...
+        """Push and pop items in FIFO order."""
+        q = StageQueue()
+        item1 = WorkItem(repo="repo1", kind=ItemKind.REPO)
+        item2 = WorkItem(repo="repo2", kind=ItemKind.ISSUE, issue=1)
+
+        q.push(item1)
+        q.push(item2)
+
+        assert len(q) == 2
+        assert q.pop() == item1
```
mvillmow: Side-effecting call inside `assert`: `assert q.pop() == item1`. Under Python `-O`, assert statements are elided and the `pop()` call never executes, leaving the queue non-empty and silently skipping the FIFO order check. Extract to a local variable:
```python
first = q.pop()
assert first == item1
```
Same fix needed on line 34 for `item2`. This was flagged in prior review (PRRT_kwDOQww0as6OXVaW / PRRT_kwDOQww0as6OXVaX) and remains unaddressed.

tests/unit/automation/pipeline/test_queues.py
```diff
@@ -0,0 +1,107 @@
...
+        assert q.pop() == item2
```
mvillmow: Same as line 33: `assert q.pop() == item2` has a side effect inside `assert`. Under `-O`, the second `pop()` is also silently skipped, leaving 2 items in the queue when `len(q) == 0` is asserted on line 35 — a latent test failure in optimized mode. Fix: `second = q.pop(); assert second == item2`.

mvillmow (commented):
All issue deliverables implemented correctly; prior R1 blocker (broken and-False assertion) resolved. Two unaddressed prior-review minor findings: q.pop() inside assert in test_queues.py:33-34.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,163 @@
...
+
+# Reason → budget key consumed when that failure repeats. Mirrors the
+# architecture doc's per-stage budget assignments; unknown reasons consume
+# none and resolve purely via the "*" default.
+_REASON_BUDGET: dict[str, str | None] = {
+    "nogo": "plan_review_iter",
+    "plan_cycles_exhausted": "plan_cycles",
+    "plan_not_go": "implement",
+    "already_implementation_go_pr": "implement",
+    "agent_error": "implement",
```
mvillmow: `"agent_error": "implement"` misattributes the budget. `agent_error` is a PR_REVIEW fail-route reason (it routes to IMPLEMENTATION); the coordinator exhausts `pr_review_iter` (or `pr_review_hard`) for it, not the `implement` counter. Same issue on line 48: `"not_implementation_go": "ci_fix"` is a CI reason that routes to PR_REVIEW, so it should consume `pr_review_iter`, not `ci_fix`. The termination proof is still sound (conservative), but the simulation doesn't model the actual budget attribution, so a future change that raises `pr_review_iter` while lowering `implement` could make these paths incorrectly cheaper in the model. Fix: map `agent_error → pr_review_iter`, `not_implementation_go → pr_review_iter`, and add `already_implementation_go_pr → None` (it routes forward to CI, not a back-edge).

hephaestus/automation/pipeline/work_item.py
```diff
@@ -0,0 +1,106 @@
...
+    branch: str = ""
+    session_ids: dict[str, str] = field(default_factory=dict)
+    labels_cache: dict[str, bool] = field(default_factory=dict)
+    payload: dict[str, Any] = field(default_factory=dict)
+    result: ItemResult | None = None
+
+    def add_history_event(self, stage: StageName, state: str, note: str = "") -> None:
+        """Record a stage transition in the history (capped at HISTORY_CAP events)."""
+        event = HistoryEvent(timestamp=_utcnow(), stage=stage, state=state, note=note)
+        self.history.append(event)
```
mvillmow, line 107: `self.history.pop(0)` is O(n) on a list. With HISTORY_CAP=200 and per-item writes this is fine in practice, but `StageQueue` already uses `deque`; using `deque(maxlen=HISTORY_CAP)` here would be O(1) and consistent. If kept as a list, at minimum document why (`deque` lacks the `list.__eq__` comparison used in tests), but the tests actually don't compare history as a whole list — they index into it — so a `deque` would work.

mvillmow (commented):
Correct and complete implementation of pipeline foundation layer. Two minor issues: _REASON_BUDGET maps agent_error/not_implementation_go to wrong budget keys (termination still holds but model is imprecise), and history.pop(0) is O(n).

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,163 @@
...
+    route = ROUTES[stage]
+    return route.fail_routes.get(reason, route.fail_routes.get("*", StageName.FINISHED))
+
+
+def _declared_budget(budget_key: str) -> int:
+    """Return the ROUTES-declared value for a budget key."""
+    return next(
+        route.budgets[budget_key] for route in ROUTES.values() if budget_key in route.budgets
+    )
+
```
mvillmow, line 76: `"no_pr": "ci_fix"` misattributes the budget. `no_pr` in CI routes directly to FINISHED (not back to CI), so it does not consume the `ci_fix` counter. The termination proof is still sound because `_fail_target(CI, "no_pr")` returns FINISHED regardless of whether `ci_fix` is incremented first — but the simulation models `ci_fix` as the back-edge guard for `no_pr`, which is wrong. Fix: `"no_pr": None` (no budget consumed; it is a terminal path, not a retry loop).

tests/unit/automation/pipeline/test_work_item.py
```diff
@@ -0,0 +1,184 @@
...
+        assert result.reason == "all stages passed"
+        assert result.final_stage == StageName.FINISHED
+
+    def test_item_result_frozen(self) -> None:
+        """ItemResult is frozen (immutable)."""
+        result = ItemResult(passed=True, reason="ok", final_stage=StageName.FINISHED)
+        with pytest.raises(AttributeError):
+            result.passed = False  # type: ignore
+
+
```
mvillmow, line 77: `datetime.now()` creates a naive (timezone-unaware) datetime for the `HistoryEvent` fixture. The `HistoryEvent` type accepts any `datetime`, but the rest of this test module (e.g. `test_work_item_timestamps_are_utc_aware`) asserts that production code always produces tz-aware datetimes. Using `datetime.now(timezone.utc)` here would keep the test fixtures consistent with the production invariant.

mvillmow (commented):
All issue deliverables land correctly: pipeline package (routing, work_item, queues, lazy __init__), tests (table-driven + hypothesis), zero-I/O AST guard, triple omit-list edit. Two minor attribution/consistency issues in tests; no blockers.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,163 @@
...
+)
+_REASONS = [*_DECLARED_REASONS, "unknown_reason"]
+
+# Reason → budget key consumed when that failure repeats. Mirrors the
+# architecture doc's per-stage budget assignments; unknown reasons consume
+# none and resolve purely via the "*" default.
+_REASON_BUDGET: dict[str, str | None] = {
+    "nogo": "plan_review_iter",
+    "plan_cycles_exhausted": "plan_cycles",
+    "plan_not_go": "implement",
```
mvillmow: `"plan_not_go": "implement"` has the same misattribution as the prior `"no_pr": "ci_fix"` issue. `plan_not_go` routes IMPLEMENTATION → PLAN_REVIEW (a fail-back exit, not an IMPLEMENTATION retry), so it does not consume the `implement` counter. The simulation terminates correctly either way (budget exhaustion just fires sooner on this code path), but the model is semantically wrong. Fix: `"plan_not_go": None` — it is a terminal escape from the IMPLEMENTATION retry loop, not a retry of it.

mvillmow (commented):
Implementation is correct and complete: all 4 production modules, 5 test files, stale omit-list triple-edit. One minor budget misattribution in property test simulation (plan_not_go attributed to implement budget); production code unaffected.

tests/unit/automation/pipeline/test_routing_properties.py
mvillmow: Fixed in b404fec: plan_not_go and not_implementation_go now map to None in _REASON_BUDGET (fail-back exits leave their stage rather than retry it, so they consume no retry budget — same class as already_implementation_go_pr and no_pr), with a comment documenting the rule. Property suite green (5 passed).

hephaestus/automation/pipeline/work_item.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    repo: str
+    kind: ItemKind
+    issue: int | None = None
+    pr: int | None = None
+    stage: StageName = StageName.REPO
+    state: str = ""
+    attempts: dict[str, int] = field(default_factory=_default_attempts)
+    history: deque[HistoryEvent] = field(default_factory=lambda: deque(maxlen=HISTORY_CAP))
```
mvillmow, line 94: `history` is typed as `deque[HistoryEvent]` but the field name and issue spec imply a list-like container. Downstream coordinators iterating or slicing history should be unaffected (`deque` supports `[0]`/`[-1]`/iteration), but callers that do `list(item.history)` or `item.history.copy()` would work while `item.history.sort()` would raise. Add a one-liner docstring note on the field or in `WorkItem`'s class docstring that history is a `deque` to make the interface explicit for future consumers (POLA).

mvillmow (commented):
All deliverables correct; prior findings resolved (q.pop() extracted, _REASON_BUDGET corrected, deque(maxlen) replaces O(n) pop(0)). One minor POLA gap: history's deque type is unannounced in class docs.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,167 @@
...
+# (implementation -> plan_review), already_implementation_go_pr (-> ci),
+# not_implementation_go (ci -> pr_review), and no_pr (-> finished) all map
+# to None.
+_REASON_BUDGET: dict[str, str | None] = {
+    "nogo": "plan_review_iter",
+    "plan_cycles_exhausted": "plan_cycles",
+    "plan_not_go": None,
+    "already_implementation_go_pr": None,
+    "agent_error": "pr_review_iter",
+    "human_blocked": "pr_review_iter",
```
mvillmow: `human_blocked` and `exhaustion` map to `pr_review_iter` here, but both route unconditionally to FINISHED — they are terminal exits, not retries. The comment at lines 38-42 explicitly states that fail-back exits that leave their stage without retrying consume no retry budget (hence `plan_not_go`, `not_implementation_go`, etc. map to `None`). The same logic applies here: assigning `pr_review_iter` inflates that counter in the simulation (a `[human_blocked, agent_error, agent_error, agent_error]` sequence exhausts the budget 1 step earlier than the real coordinator would). Fix: `"human_blocked": None, "exhaustion": None`.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,167 @@
...
+            assert budgeted_steps <= max_total_steps, (
+                "failure walk exceeded the global budget bound"
+            )
+            item.attempts[budget_key] += 1
+            if item.attempts[budget_key] > _declared_budget(budget_key):
+                item.stage = StageName.FINISHED
+                continue
+        item.stage = _fail_target(item.stage, reason)
+
+    assert item.stage in set(StageName)
```
mvillmow: `assert item.stage in set(StageName)` is trivially true — every assignment to `item.stage` in the loop body is already a valid `StageName`, so this assertion cannot fail under any input. The stated property (b) is that the walk *terminates* (reaches FINISHED), not merely that it stays valid. Consider `assert item.stage == StageName.FINISHED` to actually verify the termination claim for the budgeted-reason paths, or document that the bound is enforced by the 60-item list cap rather than the routing logic for unbudgeted reasons.

mvillmow (commented):
Product code correct and complete; two minor test-logic issues in test_routing_properties.py: _REASON_BUDGET misclassifies terminal exits as retries, and final assertion is trivially true.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,167 @@
...
+# their stage rather than retry it consume no retry budget: plan_not_go
+# (implementation -> plan_review), already_implementation_go_pr (-> ci),
+# not_implementation_go (ci -> pr_review), and no_pr (-> finished) all map
+# to None.
+_REASON_BUDGET: dict[str, str | None] = {
+    "nogo": "plan_review_iter",
+    "plan_cycles_exhausted": "plan_cycles",
+    "plan_not_go": None,
+    "already_implementation_go_pr": None,
+    "agent_error": "pr_review_iter",
```
mvillmow: `agent_error` routes PR_REVIEW → IMPLEMENTATION (`routing.py:117`), making it a fail-back exit that leaves the stage rather than retrying within it. The comment at lines 38-42 states that such exits consume no retry budget and should map to `None`. Mapping it to `"pr_review_iter"` inflates that counter on each `agent_error` transition, so a sequence like `[agent_error, agent_error, agent_error, agent_error]` exhausts the budget one step earlier than the real coordinator would — the same correctness defect that was fixed for `human_blocked` and `exhaustion`. Fix: `"agent_error": None`.

mvillmow (commented):
One correctness defect in test_routing_properties.py:48 — agent_error should map to None (fail-back exit, not a within-stage retry), same class as the human_blocked/exhaustion fix from the prior review.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,167 @@
...
+    "human_blocked": "pr_review_iter",
+    "exhaustion": "pr_review_iter",
+    "fix_exhausted": "ci_fix",
+    "not_implementation_go": None,
+    "no_pr": None,
+    "ci_red": "merge",
```
mvillmow, line 57: `"closed"` and `"timeout"` route MERGE_WAIT directly to `FINISHED` (they exit the pipeline, not retry or cycle). Charging the `merge` budget for these exits means a single `closed` or `timeout` event depletes one merge-attempt slot, so a subsequent real merge attempt may be refused earlier than intended. If these reasons mean "the PR was closed/timed out before any merge was attempted," the budget key should be `None`. Only `ci_red` (which loops back to CI and then returns to MERGE_WAIT) arguably consumes a merge slot.

tests/unit/automation/pipeline/test_routing.py
```diff
@@ -0,0 +1,313 @@
...
+            ),
+            StageName.FINISHED: Route(next=StageName.FINISHED),
+        }
+        assert expected == ROUTES
+
+
+class TestPipelineScope:
+    """Tests for PipelineScope (trimming routes to a stage subset)."""
+
+    def test_pipeline_scope_all_stages(self) -> None:
```
mvillmow, line 211: The `or stage == StageName.FINISHED` branch is dead code: `_compute_trimmed_routes` includes FINISHED when it is in `self.stages`, and `all_stages = frozenset(StageName)` includes FINISHED. The guard silently allows the test to pass even if FINISHED were accidentally excluded from the result. Tighten to `assert stage in trimmed` and add `assert StageName.FINISHED in trimmed` explicitly to close the gap.

mvillmow (commented):
Correct and complete implementation of pipeline foundation layer. Two minor test modeling gaps: closed/timeout budget attribution and a dead-code branch in all-stages scope test. No production bugs.

tests/unit/automation/pipeline/test_routing_properties.py
mvillmow: Fixed in 9a0df5c: closed/timeout now map to None in _REASON_BUDGET (pipeline exits charge no merge slot; comment contrasts ci_red which cycles for another attempt), and the all-stages scope test asserts FINISHED membership explicitly with the dead escape branch removed. 58 pipeline tests green.

tests/unit/automation/pipeline/test_routing.py
mvillmow: Fixed in 9a0df5c: closed/timeout now map to None in _REASON_BUDGET (pipeline exits charge no merge slot; comment contrasts ci_red which cycles for another attempt), and the all-stages scope test asserts FINISHED membership explicitly with the dead escape branch removed. 58 pipeline tests green.

tests/unit/automation/pipeline/test_zero_io_imports.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _forbidden(name: str) -> bool:
+    """Check if a module name is forbidden."""
+    root = name.split(".")[0]
+    return root in _FORBIDDEN_MODULES or name.startswith(_FORBIDDEN_PREFIXES)
+
+
+def test_pipeline_modules_have_zero_io_imports() -> None:
+    """Verify pipeline modules do not import I/O-related modules.
```
mvillmow, line 44: The zero-IO guard has no negative test — if `_forbidden()` were broken (e.g. returns False for everything), `test_pipeline_modules_have_zero_io_imports` would pass vacuously with an empty `violations` list. Per the `architecture-executable-convention-guard-pattern` skill (v1.4.0), guards should exercise their own failure branch. Add a test that injects a synthetic forbidden import (e.g. parse a fake `"import subprocess"` string through the AST walker) and asserts it appears in violations, confirming the guard actually catches what it claims to catch.

mvillmow (commented):
All deliverables correct; prior findings resolved (deque maxlen, q.pop() extracted, _REASON_BUDGET corrected). One minor gap: zero-IO guard lacks a negative test to verify its detection logic.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,175 @@
...
+    "already_implementation_go_pr": None,
+    "agent_error": None,
+    "human_blocked": None,
+    "exhaustion": None,
+    "fix_exhausted": "ci_fix",
+    "not_implementation_go": None,
+    "no_pr": None,
+    # ci_red cycles merge_wait -> ci for another merge attempt later, so it
+    # consumes a merge slot; closed/timeout EXIT the pipeline entirely and
+    # charge nothing (a closed PR must not deplete merge attempts).
```
mvillmow, line 56: `"fix_exhausted": "ci_fix"` maps an exit reason (CI → IMPLEMENTATION, one-way) to the `ci_fix` retry budget, which is the same structural imprecision the prior review flagged for `agent_error`. `fix_exhausted` means the CI fix budget was already exhausted upstream; charging `ci_fix` here is redundant and architecturally wrong (it models a retry of CI, not a forward exit). Since `ci_fix=1` and the exhaustion check forces FINISHED immediately, the termination proof is conservatively sound — but a future increase to `ci_fix` would expose the model mismatch. Fix: map `"fix_exhausted": None` (same as `plan_not_go`, `not_implementation_go`) since it's a forward exit, not a retry.

mvillmow (commented):
Correct and complete pipeline foundation. deque(maxlen=200) history cap is O(1) and correct; _REASON_BUDGET improved; negative guard tests added. One minor budget misattribution in test model.

tests/unit/automation/pipeline/test_routing_properties.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@given(budget_key=st.sampled_from(sorted(budget_keys())))
+def test_budget_exhaustion_is_always_finite(budget_key: str) -> None:
+    """(c) budget+1 repeats of any budgeted failure exhausts the budget.
+
+    Also proves every ROUTES budget key has a matching counter on a fresh
+    WorkItem, so no budget is untrackable.
+    """
+    item = WorkItem(repo="r", kind=ItemKind.ISSUE, issue=1)
```
mvillmow, line 175: The `if exhausted_a_budget:` guard means sequences that never hit a budgeted reason (e.g. 60 × `unknown_reason` stuck at `PR_REVIEW`'s `*`-self-loop) never assert the termination property — the test passes vacuously for those cases. The docstring says "always terminates" but the assertion only fires when a budget was actually consumed. This is correct but the invariant is weaker than advertised: consider adding a fallback assertion that the walk is bounded even for budget-free walks (e.g. `assert budgeted_steps + len([r for r in reasons if _REASON_BUDGET[r] is None]) <= max_total_steps + 60`) or noting in the docstring that termination is only asserted for budgeted walks.

mvillmow (commented):
All issue deliverables land cleanly. Budget attributions in _REASON_BUDGET are architecturally correct. One minor: termination assertion is conditional on budget exhaustion, so all-None-budget sequences exercise no invariant.

tests/unit/automation/pipeline/test_routing_properties.py
mvillmow: Fixed in bda5e14: split into an UNCONDITIONAL termination property (any budgeted-failure sequence longer than the total budget sum must land in FINISHED — asserted on every example, min_size makes vacuity impossible) plus mixed-sequence walk invariants asserted unconditionally (valid stage, global budget bound, FINISHED absorbing). Shared _drive helper; 6 property tests green.

mvillmow:
**Operator GO decision** (breaking the #1554-class review-churn deadlock):

The automation reviewer's verdict on the substance of this diff has been GO in every pass (R0-R4 of pass 1, R0-R2 of passes 2-3); each pass stranded only because it raised exactly one fresh MINOR thread per round and the 3-iteration budget converts thread-count>0 at the cap into NOGO. Every thread raised across all passes — 8 total — was addressed with a commit and resolved with cited evidence (`reopened=0` in every validator round). Independent gates:

- Strict full-coverage audit (6-dimension swarm): **GO, A- (90.3%)** with both majors remediated and CI-pinned (full-table ROUTES doc-conformance test).
- Module coverage 100% on all three pipeline modules; hypothesis properties mutation-verified (3 injected mutations killed).
- mypy whole-tree, ruff, pre-commit, and the required CI suite green.

Applying `state:implementation-go` and arming squash auto-merge per policy.

mvillmow:
Superseded by https://github.com/HomericIntelligence/ProjectHephaestus/pull/1837: identical tree (byte-for-byte, final head bda5e14) re-cut as one clean signed+DCO commit. This branch accumulated five automation-authored commits without DCO trailers — pr-policy Check 4 rightly blocks them, and history rewrite via force-push is disabled in this environment. All 8 review threads here were addressed and resolved; full provenance carried in the new PR's body.

PR #1686, opened 2026-06-30
refactor(utils): extract read_int_env helper to prevent env-var crashes

mvillmow:
## Summary
Extract the safe `_read_int_env()` helper from `claude_timeouts.py` into `hephaestus/utils/helpers.py` and replace all bare `int(os.environ.get(...))` calls with robust env-var validation. Prevents `ValueError` crashes on malformed integer environment variables and ensures consistent error handling with logging across `ci_driver.py` and `loop_runner.py`.

## Changes
- Move `_read_int_env()` helper to `hephaestus/utils/helpers.py` as a reusable public utility
- Replace bare `int(os.environ.get(...))` in `ci_driver.py:1425` (HEPH_PR_MERGE_MAX_WAIT) with safe helper
- Replace bare `int()` in `loop_runner.py:952` (HEPHAESTUS_RATE_GUARD_THRESHOLD) with safe helper
- Update `claude_timeouts.py` to import and use the extracted helper
- Add unit tests for `read_int_env()` error handling and logging

## Testing
- Unit tests verify graceful handling of non-numeric env-var input with logging
- Verify ci_driver.py and loop_runner.py no longer crash on malformed env vars
- Confirm all env-var reads use the same safe helper

## Closes
Closes #1431

Generated by Claude Code via ProjectHephaestus automation.

PR #1836, opened 2026-07-04
test(automation): cover timeout env fallbacks

### LINKED ISSUE #1429: [fix] ci_driver.py and loop_runner.py bare int(os.environ.get()) — use _read_int_env from claude_timeouts
## Motivation

`ci_driver.py:1425` reads `HEPH_PR_MERGE_MAX_WAIT` with `int(os.environ.get(..., "1800"))` which will raise `ValueError` on non-numeric input. Meanwhile, `claude_timeouts.py` has a robust `_read_int_env()` helper that handles this gracefully with logging.

Similarly, `loop_runner.py:952` reads `HEPHAESTUS_RATE_GUARD_THRESHOLD` with bare `int()` — same crash risk.

The `claude_timeouts.py:_read_int_env()` function already solves this pattern. It should be reused everywhere an integer env var is read.

## Proposed Solution

1. Make `_read_int_env` a public utility (move to `hephaestus/utils/helpers.py` or keep in `claude_timeouts.py` and export)
2. Replace all bare `int(os.environ.get(...))` calls with the safe helper
3. Add validation logging like `claude_timeouts.py` already does

## Impact

- Prevents runtime crashes on malformed env vars
- Consistent error handling across all env var reads
- Very low risk

## Files affected

- `hephaestus/automation/claude_timeouts.py` (export `_read_int_env`)
- `hephaestus/automation/ci_driver.py` (line 1425)
- `hephaestus/automation/loop_runner.py` (line 952)
- Any other bare `int(os.environ.get(...))` calls

mvillmow:
## Summary
- Adds regression coverage for safe timeout environment parsing across CI-driver, loop-runner, constants, and helper timeout configuration.
- Preserves the current main-branch runtime helper implementation while proving malformed integer overrides fall back instead of crashing.

## Test Plan
- [x] `env PIXI_CACHE_DIR=/tmp/pixi-cache-heph-1429 pixi run pytest tests/unit -v`
- [x] `env PIXI_CACHE_DIR=/tmp/pixi-cache-heph-1429 pixi run python -m mypy hephaestus/`
- [x] `env PIXI_CACHE_DIR=/tmp/pixi-cache-heph-1429 pixi run ruff check hephaestus/ tests/`
- [x] `env PIXI_CACHE_DIR=/tmp/pixi-cache-heph-1429 pixi run ruff format --check hephaestus/ tests/`
- [x] `env PIXI_CACHE_DIR=/tmp/pixi-cache-heph-1429 PRE_COMMIT_HOME=/tmp/pre-commit-heph-1429 pixi run pre-commit run --files tests/unit/automation/test_ci_driver.py tests/unit/automation/test_loop_runner.py tests/unit/constants/test_constants.py tests/unit/utils/test_helpers.py`

Closes #1429

Implemented-By: Codex
Co-Authored-By: Codex <noreply@openai.com>


tests/unit/automation/test_ci_driver.py
```diff
@@ -2706,6 +2706,24 @@ def test_open_and_green_times_out(
...
+            patch.object(
+                driver,
+                "_gh_pr_state",
+                return_value={"state": "MERGED"},
+            ),
+            patch("hephaestus.automation.ci_driver.time.sleep"),
+        ):
+            # A MERGED PR returns immediately; the point is the env read no
+            # longer crashes before the poll loop begins.
+            assert driver._wait_for_pr_terminal(1, 2) == "MERGED"
```
mvillmow, line 2725: The test proves the env read doesn't crash (no ValueError), but because the PR state is MERGED the `max_wait` budget is never consumed. Consider adding a separate parametrize case or inline assertion that `read_timeout_env("HEPH_PR_MERGE_MAX_WAIT", 1800)` returns 1800 when the env var is malformed — directly on `constants.read_timeout_env` in `test_constants.py` — to lock in the actual fallback value, not just the absence of a crash.

mvillmow (commented):
Tests-only PR adding regression coverage for #1429's env-read fixes (already on main via #1686). All four test additions are behaviorally correct; one minor gap: ci_driver test doesn't verify the fallback value, only absence of crash.

tests/unit/utils/test_helpers_timeouts.py
```diff
@@ -44,3 +44,14 @@ def test_network_timeout_reads_valid_override(monkeypatch: pytest.MonkeyPatch) -
     finally:
         monkeypatch.delenv("HEPHAESTUS_SUBPROCESS_NETWORK_TIMEOUT", raising=False)
         importlib.reload(helpers)
+
+
+def test_metadata_timeout_reads_valid_override(monkeypatch: pytest.MonkeyPatch) -> None:
+    """A valid integer METADATA override is honoured on reimport."""
```
mvillmow, line 50: This test covers a valid override but the plan (issue #1429) also called for a malformed-env regression: `HEPHAESTUS_SUBPROCESS_METADATA_TIMEOUT=bad_value` should fall back to 10 without crashing. Consider adding a parallel `test_metadata_timeout_falls_back_on_malformed_value` (same `importlib.reload` pattern) to close that gap.

mvillmow (commented):
Test-only PR adding correct crash-safety regressions for ci_driver/loop_runner/constants; minor gap: no malformed-env fallback test for METADATA_TIMEOUT in helpers.

mvillmow (commented):
Tests-only increment adds malformed-env regression coverage across four files; all tests are semantically correct and cover the required fallback paths.

Issue #1855, opened 2026-07-06
Epic: harden the #1809 automation pipeline (strict-review findings)


mvillmow:
# Epic: harden the #1809 automation pipeline (strict-review findings)

**Type:** Epic / hardening. **Do not implement this issue directly** — each finding is filed as its own sub-issue (linked below). This epic tracks remediation of the strict review of the merged pipeline surface.

## Context

Epic #1809 remapped the automation loop into a queue-based, in-process state-machine pipeline. Sub-issues #1810–#1817 are merged: the full `hephaestus/automation/pipeline/` package (coordinator, worker pool, seeding, admission, 9 stages, `pipeline_github.py`, summary — ~9.3k LoC, behind the `--pipeline` flag, default off). The cutover (#1818) and cleanup wave (#1819–#1823) remain open.

A 7-agent strict review audited the **merged** implementation against `origin/main`. Findings were synthesized, de-duplicated, and verified against the code. This epic groups **78 filed issues**: **3 CRITICAL, 16 MAJOR, 26 MINOR, 33 NIT**.

One not-real candidate was dropped (`CompletionQueue` is unbounded, so its `put` cannot block). Auto-follow-up #1850's two items were already fixed on `origin/main` (#1850 closed). #1847 (prompt-too-long) was already filed.

## The three CRITICAL findings (headline defects)

- **C1 — pr_review GO gate is not severity-aware** (`stages/pr_review.py:703`). A GO stands only when `automation_unresolved + human_unresolved == 0`, no severity filter. A MINOR automation thread the reviewer itself GO'd is downgraded to NOGO every round; the address leg can't force-resolve a nit the agent considers fine, so the item burns all rounds → `state:skip`. The #1554 minor-thread deadlock, re-introduced.
- **C2 — the re-plan cycle can never converge** (`stages/planning.py:287`, contract `stages/base.py:129`). VERIFY advances on `has_existing_plan`, whose mandated contract reuses labels-first `is_plan_review_go`, which returns False while `state:plan-no-go` is present. On the `plan_cycles` re-plan path plan_review has just applied that label, so VERIFY never advances however good the fresh plan — it RETRYs until the `plan` budget exhausts and `FINISH_FAIL`s. `plan_cycles=2` is dead.
[...]

PR #1937, opened 2026-07-06
fix(automation): key repo-scoped label cache with a lock guard

### LINKED ISSUE #1858: [crit] Cross-repo unlocked module-level label cache corrupts durable journal writes under the multithreaded coordinator
`PipelineGitHub._label_names()` org-scoped path (pipeline_github.py:192) returns `github_api.gh_list_labels()`, which reads/writes an unguarded module global `_api._label_cache` (github_api/labels.py:24-31) that is neither repo-keyed nor lock-protected. With the coordinator now running stages multi-threaded across repos, thread A seeding repo X populates the cache with X's labels; thread B for repo Y then sees X's cache, concludes Y's `state:*` labels already exist, and skips `_create_label` in `_add_labels` (pipeline_github.py:204-215) -- so `issue edit --add-label` fails or applies against a repo missing the label. Because these are the durable journal writes, a dropped one means a crash-restart re-seed lands at the WRONG stage (e.g. an epic never tagged `state:skip` -> re-queued forever), breaking the durable-write-precedes-push invariant. Fix: key the cache by repo slug and guard with a lock, or make the org path fetch fresh per-repo like the repo-scoped branch already does.

**Location:** `hephaestus/automation/pipeline_github.py:192`
**Severity:** CRITICAL  
**Category:** `shared-cache-concurrency`

Found by the epic #1809 strict review of the merged pipeline surface (reviewed against `origin/main`).

Related to #1855
Part of epic #1809

mvillmow:
Fixes the cross-repo label-cache corruption (#1858, a CRITICAL from the epic #1809 strict review).

The module-global `_label_cache` in `github_api.labels` was unkeyed and unguarded, so under the multithreaded pipeline coordinator one repo could mask another repo state:* labels, dropping durable journal writes and breaking the crash-restart invariant.

- Introduce `ThreadSafeCache` (lock-protected, keyed) in `hephaestus/utils/cache.py` (documented in COMPATIBILITY.md).
- Key the label cache by `(org, repo)` and guard all access; org-scoped path fetches fresh per repo.

Part of the #1855 hardening sub-epic.

Closes #1858

### DIFF
--- a/hephaestus/utils/cache.py
+++ b/hephaestus/utils/cache.py
@@ -68,9 +68,11 @@ class ThreadSafeCache(Generic[K, V]):
     def add_to_entry(self, key: K, item: Any) -> None:
         """If *key* holds a set entry, add *item* to it under the lock.
 
-        No-op when the key is absent or expired — mirrors the label cache's
-        "only augment an already-populated entry" contract. TTL timestamp is
-        preserved (augmenting does not refresh expiry).
+        No-op when the key is absent — mirrors the label cache's "only
+        augment an already-populated entry" contract. This does not check TTL
+        expiry, so an entry past its TTL is still augmented; the stale
+        augmented entry is replaced on the next ``get_or_compute`` call. TTL
+        timestamp is preserved (augmenting does not refresh expiry).
 
         Intended for use with ``ThreadSafeCache[K, set[V]]`` where adding an
         element to a cached set is atomic (label cache in github_api).
--- a/hephaestus/utils/cache.py
+++ b/hephaestus/utils/cache.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import threading
 import time
 from collections.abc import Callable
-from typing import Generic, TypeVar
+from typing import Any, Generic, TypeVar
 
 K = TypeVar("K")
 V = TypeVar("V")
@@ -64,3 +64,24 @@ class ThreadSafeCache(Generic[K, V]):
         """Remove all entries. For test isolation and long-lived processes."""
         with self._lock:
             self._cache.clear()
+
+    def add_to_entry(self, key: K, item: Any) -> None:
+        """If *key* holds a set entry, add *item* to it under the lock.
+
+        No-op when the key is absent or expired — mirrors the label cache's
+        "only augment an already-populated entry" contract. TTL timestamp is
+        preserved (augmenting does not refresh expiry).
+
+        Intended for use with ``ThreadSafeCache[K, set[V]]`` where adding an
+        element to a cached set is atomic (label cache in github_api).
+        """
+        with self._lock:
+            entry = self._cache.get(key)
+            if entry is not None:
+                value, ts = entry
+                self._cache[key] = ({*value, item}, ts)  # type: ignore[assignment,misc]
+
+    def remove(self, key: K) -> None:
+        """Remove *key* from the cache. No-op if the key is absent."""
+        with self._lock:
+            self._cache.pop(key, None)
--- a/tests/unit/automation/test_github_api.py
+++ b/tests/unit/automation/test_github_api.py
@@ -1112,7 +1112,7 @@ class TestGhListLabels:
 
     def setup_method(self) -> None:
         """Reset module-level label cache before each test."""
-        _github_api_module._label_cache = None
+        _github_api_module._label_cache.clear()
 
     @patch("hephaestus.automation.github_api._gh_call")
     def test_returns_set_of_label_names(self, mock_gh_call: Any) -> None:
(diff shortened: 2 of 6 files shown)


tests/unit/utils/test_cache.py
```diff
@@ -152,3 +152,37 @@ def worker(i: int) -> None:
...
     observed = {i: cache.get_or_compute(i, lambda: -1) for i in range(num_threads)}
     assert observed == {i: i for i in range(num_threads)}
+
+
+def test_add_to_entry_augments_existing_set() -> None:
+    """add_to_entry adds an item to an existing set-typed entry."""
+    c: ThreadSafeCache[str, set[str]] = ThreadSafeCache()
+    c.get_or_compute("k", lambda: {"a"})
+    c.add_to_entry("k", "b")
+    assert c.get_or_compute("k", lambda: set()) == {"a", "b"}
```
mvillmow: Outstanding prior review finding (PRRT_kwDOQww0as6OgaQV): `lambda: set()` is an unnecessary wrapper — replace with `set` directly. The lambda has no argument transformation; passing the callable directly is idiomatic and was the explicit fix requested.

hephaestus/utils/cache.py
```diff
@@ -64,3 +64,24 @@ def clear(self) -> None:
...
+        No-op when the key is absent or expired — mirrors the label cache's
+        "only augment an already-populated entry" contract. TTL timestamp is
+        preserved (augmenting does not refresh expiry).
+
+        Intended for use with ``ThreadSafeCache[K, set[V]]`` where adding an
+        element to a cached set is atomic (label cache in github_api).
+        """
+        with self._lock:
+            entry = self._cache.get(key)
+            if entry is not None:
```
mvillmow, line 82: The docstring at line 71 promises `add_to_entry` is a no-op 'when the key is absent or expired', but the implementation checks only `entry is not None` — it does NOT check TTL expiry. An entry past its TTL will still be augmented, inconsistent with `get_or_compute`'s TTL check (`time.monotonic() - ts < self._ttl`). Practical impact is low (the stale augmented entry is replaced on the next `get_or_compute` call), but the contract is broken. Either add the TTL check here (consistent with `get_or_compute`) or fix the docstring to say 'No-op when the key is absent'.

mvillmow (commented):
Core fix correct: repo-keyed ThreadSafeCache with per-key refresh eviction resolves #1858. Two minor issues: outstanding lambda:set() finding still present in test_cache.py:162, and add_to_entry TTL doc/impl mismatch in cache.py:80.

tests/unit/utils/test_cache.py
mvillmow: `c.get_or_compute("k", set)` passes `set` (zero-arg callable) as the compute factory. On a cache hit it is never called, so the test passes vacuously for the wrong reason. A reader skimming the factory arg sees `set` and expects it to produce the augmented `{"a", "b"}`, but it would actually produce `set()` if the cache missed. Prefer `lambda: set()` (or any sentinel that would fail visibly on an unexpected miss) to make the test's intent clear.

mvillmow (commented):
Correct repo-keyed, lock-protected label cache via ThreadSafeCache. refresh=True now does keyed eviction (not clear). All test touchpoints migrated. One minor misleading compute-arg in test_cache.py:162.

tests/unit/automation/test_github_api.py
```diff
@@ -1161,22 +1161,90 @@ def test_create_label_calls_gh(self, mock_gh_call: Any) -> None:
...
+        def _run(slug: str) -> None:
+            t = threading.current_thread()
+            t.name = slug
+            results[slug] = gh_list_labels(refresh=True)
+
+        with patch(
+            "hephaestus.automation.github_api.labels._label_cache_key", side_effect=mock_key_fn
+        ):
+            with patch("hephaestus.automation.github_api._gh_call", side_effect=mock_gh_call):
+                t1 = threading.Thread(target=_run, args=("t1",), name="t1")
```
mvillmow, line 1231: The concurrency test spawns two threads but has no `threading.Barrier` to force them to overlap in `gh_list_labels`. Without a barrier, thread t1 may complete entirely before t2 starts, so the test never exercises concurrent cache access — it only verifies sequential isolation (which `test_distinct_repos_do_not_share_cache` already covers). Add a `threading.Barrier(2)` that both threads wait on inside `_run` before calling `gh_list_labels`, matching the plan's original design and the `architecture-metaclass-threadlocal-thread-safety` skill pattern.

mvillmow (commented):
Correct per-key repo-slug cache with lock guard; refresh now does per-key eviction (P7 resolved); minor: concurrency test lacks barrier so threads may not actually overlap.

mvillmow (commented):
All prior findings addressed: per-key refresh via remove(key), honest TTL docstring, barrier-enforced concurrent test, sentinel-based cache-hit assertion. LGTM.

Issue #2057, opened 2026-07-10
fix(pipeline): duplicate-issue assert in _drain_implementation crashes entire org run


mvillmow:
## Problem

`_drain_implementation` (coordinator.py) hard-`assert`s that the IMPLEMENTATION queue holds no duplicate issue numbers (added in #1952 / closing #1879). In a live `hephaestus-automation-loop --org` run this assertion fired on `Odysseus #71, #206`, raised through `run()`, and **crashed the entire org-wide pipeline loop** — dropping 321 pending items across 15 repos (`dispositions: {'pass': 16, 'pending': 321}`).

```
File ".../pipeline/coordinator.py", line 817, in _drain_implementation
    assert not duplicate_issues, (  # noqa: S101
AssertionError: implementation queue must not contain duplicate issue numbers: [71, 206]
```

## Root cause

A transient retry / fail-back (`_route_retry` → `_push_item(item, item.stage, ...)`) can re-enqueue an issue into IMPLEMENTATION while a prior copy of that issue is still queued, so the queue may briefly hold two work items for the same issue number. The `issue_items` dict build (#1879) requires uniqueness, but the chosen guard was a fatal `assert` rather than defensive handling — strictly more fragile than the silent-collapse it replaced.

## Fix

Collapse duplicates defensively at the top of `_drain_implementation`: keep the first-queued item, terminalize later duplicates as `superseded` (pass), log a warning, and dispatch normally. Never crash the run.

Regression test updated: `test_duplicate_issue_numbers_collapse_to_first_queued`.

PR #2061, opened 2026-07-11
[Fix] Reap worker pool on fatal-exception exit

### LINKED ISSUE #2059: fix(pipeline): coordinator fatal-exception teardown leaks in-flight AgentJob subprocesses
## Problem

When `PipelineCoordinator.run()` exits via a **fatal exception** (not a signal), in-flight `AgentJob` subprocesses (e.g. `claude` PR-reviewers) are **leaked** — they keep running after the run has logically ended.

Observed live: after the duplicate-issue assertion crash (#2057), `run_end` fired with `exit_code 1` and the summary printed, but a `claude --name ProjectOdyssey_5545_pr-reviewer_...` child kept running for ~19 minutes, holding the main process open until it happened to finish.

## Root cause

The worker pool is only shut down on the **signal path**:

- `coordinator.py:1431` — `self.pool.shutdown()` is called **only** inside `_teardown_immediate()`, which runs on SIGINT/SIGTERM (`self._immediate`).
- The `run()` `finally` block (`coordinator.py:515-537`) calls `_finalize_resumable()`, which **early-returns unless `self.shutdown.is_set()`** (`coordinator.py:1441`). A fatal `Exception` sets `self._fatal = True` but does **not** set `self.shutdown`.

Net: on a fatal exception, neither `pool.shutdown()` nor in-flight reaping runs. Any submitted-but-unfinished AgentJob subprocess leaks.

## Proposed fix

In the `run()` `finally` block, ensure the worker pool is drained/cancelled and in-flight items are parked RESUMABLE regardless of *why* the loop exited:

- Call `self.pool.shutdown()` on the fatal-exit path too (idempotent / guarded), and
- Have `_finalize_resumable()` also finalize when `self._fatal` is set (not just `self.shutdown.is_set()`), so in-flight items are terminalized/parked and their subprocesses reaped.

## Tests

- Unit: simulate a stage raising inside `run()`; assert `pool.shutdown()` is called exactly once and in-flight items are parked RESUMABLE on the fatal path.
- Guard against double-shutdown when a signal and a fatal exception coincide.

## Context

Surfaced while fixing #2057 (the crash that triggered this leak). Independent of #2057 — this hardens teardown for *any* fatal exit.

mvillmow:
## Summary

`PipelineCoordinator.run()` only shut the worker pool down on the **signal** path (`_teardown_immediate`, gated on `self.shutdown`). A **fatal exception** sets `self._fatal` but never `self.shutdown`, so the `finally` block's `_finalize_resumable()` early-returned and `pool.shutdown()` was never called — **leaking the executor and any in-flight `AgentJob` subprocesses** (e.g. `claude` reviewers).

Observed live: after the #2057 crash, `run_end` fired with `exit_code 1` but a `claude` PR-reviewer child kept running ~19 minutes, holding the process open until it happened to finish.

## Change

- Extract an idempotent `_shutdown_pool()` (guarded by `_pool_shut_down`) that cancels the pool and parks in-flight items RESUMABLE.
- Call it from `run()`'s `finally` on **every** exit path (`_teardown_immediate` also routes through it).
- `_finalize_resumable()` now also runs when `self._fatal` is set.
- Crucially, the fatal path does **not** set `self.shutdown`, so `_exit_code()` still reports `1` (fatal), not `130` (interrupt).

## Testing (TDD)

- `test_fatal_exception_shuts_down_pool_and_parks_in_flight` — verified RED without the fix (pool never reaped), GREEN with it. Asserts pool shut down exactly once, in-flight maps cleared, in-flight item parked RESUMABLE, `shutdown` NOT set, exit code `1`.
- `test_signal_teardown_shuts_down_pool_once` — signal path stays single-shutdown (idempotency guard against signal-then-`finally`).
- `pixi run pytest tests/unit/automation/pipeline/` — 838 passed.
- `pixi run mypy` clean (544 files); `ruff` + pre-commit clean on changed files.

Follow-up from #2057 (which fixed the crash that surfaced this leak). Independent — hardens teardown for any fatal exit.

Closes #2059

### DIFF
--- a/hephaestus/automation/claude_invoke.py
+++ b/hephaestus/automation/claude_invoke.py
@@ -23,15 +23,18 @@ What lives here:
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import os
 import re
+import signal
 import subprocess
 import uuid
 from dataclasses import dataclass
 from pathlib import Path
 
+from hephaestus.automation import subprocess_registry
 from hephaestus.automation.agent_config import (
     agent_default_timeout,
     fallback_model,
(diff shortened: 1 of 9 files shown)


mvillmow:
Automated PR review result: GO.

No unresolved blocking review threads were found by the automation reviewer. The pipeline marked this PR `state:implementation-go` and armed auto-merge.

Issue #2107, opened 2026-07-12
fix(pipeline): add upstream idempotency guard so duplicates never reach the IMPLEMENTATION queue


mvillmow:
## Context

Follow-up from #2057 / PR #2058. That PR added a **drain-level safety net**: `_drain_implementation` now collapses transient duplicate work items (same `(repo, issue)`) instead of crashing on a hard assert, and dispatches cross-repo same-number items as distinct work.

That is defense-in-depth. The **root cause** remains: a duplicate `(repo, issue)` work item can be enqueued into the IMPLEMENTATION queue in the first place — most likely by re-seeding (`_seed_pass` / `_seed_products`) an issue while a prior copy is still queued or in-flight. Neither path has an "already queued / already in-flight" idempotency guard (grep for `seen` / `in_flight` at seed time returns nothing).

## Ask

Add an upstream idempotency guard so a `(repo, issue)` already present in a stage queue or in `in_flight` is not re-enqueued. The drain-level net from #2058 stays as belt-and-suspenders.

## Acceptance criteria

- Seeding / re-push refuses to enqueue a `(repo, issue)` that is already queued (any stage) or in-flight.
- Unit test: re-seed while a copy is queued → no second queue entry (the drain-level dedup warning should NOT fire, proving the dup never entered).
- The `_drain_implementation` dedup remains as a safety net (do not remove it).

## Evidence

Surfaced by the strict PR-alignment review of #2058 (finding #2, MAJOR: "symptom not root cause").

mvillmow:
# Implementation Plan


Confirmed. The fix: amend this existing test to inject the drain-net collision *directly* (bypassing the upstream guard, since the guard now prevents this collision from arising through `_push_item`). Pushing straight onto the queue via `queues[...].push()` mirrors how the drain re-pushes deferred items (`coordinator.py:866`) and models the belt-and-suspenders scenario the guard can't fully eliminate (e.g. a future/unforeseen path). I now have everything for the revised plan.

# Implementation Plan

## Objective

[...]

PR #2119, opened 2026-07-12
fix(tests): stop upstream guard from masking drain-level dedup test (main red)

### LINKED ISSUE #2118: test: main red — test_three_duplicates_collapse_to_first fails after upstream idempotency guard (#2109) shadows drain-level dedup
## Summary

`main` is RED. `tests/unit/automation/pipeline/test_coordinator.py::TestImplementationAdmission::test_three_duplicates_collapse_to_first` fails on current main:

```
AssertionError: assert <StageName.IMPLEMENTATION> is <StageName.FINISHED>
  where <StageName.IMPLEMENTATION> = WorkItem(repo='repo-a', ..., issue=21, stage=<StageName.IMPLEMENTATION>).stage
```

## Root cause

The upstream idempotency guard added in #2107/#2109 (`coordinator.py:1195-1201`) drops a duplicate `(repo, issue)` at `_push_item` time — it logs "seed skipped" and returns before enqueueing. The stale test pushes 3 copies of issue #21 via `_push_item`, expecting dup2/dup3 to flow into `_drain_implementation` and be terminalized to `FINISHED` (superseded) per the #2057 drain-level dedup contract. With the guard in place, dup2/dup3 are dropped at push and never reach the drain, so their stage stays `IMPLEMENTATION`.

The guard's own PR updated the *sibling* test (`test_duplicate_in_implementation_queue_terminalized_not_fatal`) to inject duplicates BELOW the guard (straight onto the queue) so it still exercises the drain-level net — but `test_three_duplicates_collapse_to_first` was missed.

## Fix

Apply the same below-the-guard injection pattern to `test_three_duplicates_collapse_to_first` so it continues to test the drain-level defense-in-depth 3-way collapse (#2057) rather than the guard that now shadows it. The guard behavior itself is correct; only the test needed updating.

mvillmow:
Fixes main being RED.

`test_three_duplicates_collapse_to_first` began failing once the upstream idempotency guard (#2107/#2109) started dropping duplicate `(repo, issue)` keys at `_push_item` time — the test still pushed 3 duplicates via `_push_item` expecting them to reach `_drain_implementation` and be terminalized, but the guard now drops dup2/dup3 before they enqueue.

This applies the same below-the-guard injection the guard PR already applied to the sibling test, so the drain-level dedup (#2057) is still exercised. The guard behavior is correct; only the stale test needed updating.

Closes #2118

### DIFF
--- a/tests/unit/automation/pipeline/test_coordinator.py
+++ b/tests/unit/automation/pipeline/test_coordinator.py
@@ -606,8 +606,17 @@ class TestImplementationAdmission:
         first = _issue_item(21, StageName.IMPLEMENTATION)
         dup2 = _issue_item(21, StageName.IMPLEMENTATION)
         dup3 = _issue_item(21, StageName.IMPLEMENTATION)
+        # Inject the collisions BELOW the upstream idempotency guard (#2107):
+        # push straight onto the queue (the same raw API the drain uses to
+        # re-push deferred items, coordinator.py:873) so this test exercises
+        # the drain-level safety net for duplicates that arrived by a path the
+        # upstream guard does not cover. (The guard would otherwise drop dup2/
+        # dup3 at push time, so they'd never reach the drain-level collapse.)
         for it in (first, dup2, dup3):
-            coordinator._push_item(it, StageName.IMPLEMENTATION, enter=True)
+            it.payload["_enter_pending"] = True
+            coordinator._seen_item_ids.add(id(it))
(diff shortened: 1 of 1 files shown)


mvillmow:
Automated PR review result: GO.

No unresolved blocking review threads were found by the automation reviewer. The pipeline marked this PR `state:implementation-go` and armed auto-merge.

PR #2295, opened 2026-07-18
[Fix] Route unmapped-stage items to finished(fail) instead of KeyError

### LINKED ISSUE #2294: Coordinator._route crashes with KeyError on a poisoned REPO-stage item (no fallback route)
Observed live: a `--phases drive-green` run poisoned the repo/seeding item, then crashed:
```
ERROR coordinator: pipeline item Hephaestus poisoned at repo; routing to finished(fail)
Traceback ... coordinator.py:1127 _run_item -> self._route(item, result)
coordinator.py:1208 _route -> route = self._routes[item.stage]
KeyError: <StageName.REPO: 'repo'>
```
`_route` indexes `self._routes[item.stage]` with no fallback, so a poisoned REPO-stage item (or any stage absent from the scoped `_routes` map, e.g. under `--phases drive-green`) raises KeyError instead of routing to finished(fail). The poison handler that just logged "routing to finished(fail)" cannot actually reach FINISHED.

Fix: `_route` should default to a terminal finished(fail) route when `item.stage` is not in `_routes` (or REPO should always have a terminal route). Add a regression test that a poisoned REPO item under a partial `--phases` scope routes to finished(fail) rather than raising.

Related: the repo/seeding stage's <15s stall contract is unrealistic against a large state:skip backlog (observed 154.4s), which is what poisoned the item — consider raising the repo-stage budget or making seeding incremental.

mvillmow:
## Summary

A `--phases drive-green` run crashed at the seeding stage:
```
ERROR coordinator: pipeline item Hephaestus poisoned at repo; routing to finished(fail)
Traceback ... coordinator._route -> route = self._routes[item.stage]
KeyError: <StageName.REPO: 'repo'>
```
`_route` indexed `self._routes[item.stage]` directly. Under a partial `--phases` scope, `trimmed_routes` omits `StageName.REPO` even though the seeder always pushes a REPO item — so a poisoned REPO item raised `KeyError` and killed the entire run (the poison handler that logged "routing to finished(fail)" could never reach the sink).

## Fix

- `_route` now routes any stage absent from the run's route table to `finished(fail)` — failing closed to the sink that the poison path already intended, instead of crashing.
- The unscoped path now uses `dict(ROUTES)` (a copy) rather than aliasing the module-level `ROUTES`, so an in-place edit of `self._routes` can never corrupt the shared table for other runs/tests.

## Tests

New `test_stage_absent_from_route_table_finishes_failed` pins the fallback (a poisoned REPO item with REPO removed from `_routes` finishes failed, no KeyError). Updated `test_full_run_uses_global_routes` to the copy contract. 951 pipeline tests pass; lint/mypy clean.

Closes #2294

### DIFF
--- a/hephaestus/automation/pipeline/coordinator.py
+++ b/hephaestus/automation/pipeline/coordinator.py
@@ -413,7 +413,10 @@ class Coordinator:
             self._routes = config.scope.trimmed_routes()
             self._routes.setdefault(StageName.FINISHED, ROUTES[StageName.FINISHED])
         else:
-            self._routes = ROUTES
+            # Copy, not alias: ``ROUTES`` is a module-level shared table, so an
+            # accidental in-place edit of ``self._routes`` would corrupt every
+            # other run/test. The table is small and built once per run.
+            self._routes = dict(ROUTES)
 
         self._install_signals = install_signals
         self._seq = 0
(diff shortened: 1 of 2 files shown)


mvillmow:
✅ /athena:pr-review GO (independent swarm review): fallback correctly fails closed to FINISHED, cannot mis-finish a routable stage, dict(ROUTES) copy is behavior-preserving + defensive, both tests fail-on-break. All required checks green.

PR #2310, opened 2026-07-18
[Fix] Load default prompt templates by __file__ path, not PackageLoader

### LINKED ISSUE #2308: PackageLoader('hephaestus.prompts') races the editable-install rebuild — use a __file__-relative loader
## Observed (2026-07-18 06:40 run)
The loop's launch rebuilt the editable install (`Building… Uninstalled 1… Installed 1`), then within a ~5-minute window (06:44–06:49) **12 pr_review jobs crashed** identically:
```
ValueError: PackageLoader could not find a 'templates/default' directory in the 'hephaestus.prompts' package.
  hephaestus/prompts/catalog.py:42 __init__ -> jinja2 PackageLoader
```
The traceback imports from the main checkout (templates ARE present on disk), and the crashes stop after 06:49 — clean GOs from 07:08 on. So it is a **startup import/packaging race**: workers imported `hephaestus.prompts` before the freshly-rebuilt editable install's package-data / importlib metadata settled. `PackageLoader` resolves `templates/default` via importlib package metadata, which is exactly what's transiently inconsistent right after `uv sync`.

## Impact
Each crash is read by the reviewer as "no verdict," cascading into "3 consecutive reviewer failures — failing back to implementation" (10×), which regresses those PRs to re-implementation and turns a driving run into a new-work run (see the sibling file-overlap-livelock issue).

## Fix
`catalog.py:42` — replace `PackageLoader("hephaestus.prompts", "templates/default")` with a `FileSystemLoader(Path(__file__).parent / "templates" / "default")` (path-relative, no importlib-metadata dependency, immune to the post-rebuild race). Alternatively/additionally, add the startup preflight (#2283) that renders one prompt in the parent before spawning workers so a transient miss fails fast rather than torching ~13 review jobs.

mvillmow:
## Summary

`PackageLoader("hephaestus.prompts", "templates/default")` (`catalog.py:42`) resolves the template tree via importlib package metadata, which is **transiently inconsistent for a few seconds after an editable-install rebuild**. The automation loop's launch runs `uv sync` (rebuild) and immediately spawns workers; those that import the catalog inside that window crashed with `PackageLoader could not find 'templates/default'` **even though the files are on disk** — 12 reviewer jobs died this way in one run (06:44–06:49), each misread by the pipeline as a "reviewer failure" and regressed to implementation.

## Fix

Resolve the packaged defaults with a `__file__`-relative `FileSystemLoader` (`_DEFAULT_TEMPLATES_DIR = Path(__file__).resolve().parent / "templates" / "default"`) — no importlib-metadata dependency, immune to the rebuild race. The harness-override `FileSystemLoader` path is unchanged.

## Verification

The existing byte-hash compatibility test (`test_default_catalog_render_preserves_planning_compatibility`) confirms rendered output is **unchanged** (same SHA256). New `test_default_templates_resolve_by_filesystem_path_not_package_metadata` pins the metadata-independent resolution. 92 prompt tests pass; lint/format/mypy clean.

Closes #2308

### DIFF
--- a/hephaestus/prompts/catalog.py
+++ b/hephaestus/prompts/catalog.py
@@ -13,9 +13,13 @@ from jinja2 import (
     ChoiceLoader,
     Environment,
     FileSystemLoader,
-    PackageLoader,
     StrictUndefined,
 )
+
+#: Packaged default templates, resolved by filesystem path relative to this
+#: module so loading never depends on importlib package metadata (which races
+#: an editable-install rebuild, #2308).
+_DEFAULT_TEMPLATES_DIR = Path(__file__).resolve().parent / "templates" / "default"
 
 _ACTIVE_CATALOG: ContextVar[PromptCatalog | None] = ContextVar(
     "hephaestus_active_prompt_catalog", default=None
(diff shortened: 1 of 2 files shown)


mvillmow:
✅ /athena:pr-review GO (independent swarm review, Closes #2308): __file__-relative FileSystemLoader replaces PackageLoader; byte-hash test proves output unchanged; metadata-independence pinned. Required checks green. Queued.

PR #2468, opened 2026-07-26
[nit] Clarify root-level metadata navigation

### LINKED ISSUE #2183: [nit] Clarify root-level metadata navigation
Section 1 NITPICK. The root has many purposeful files; improve top-level navigation so discoverability does not depend on stale duplicated documentation.

mvillmow:
## Summary
Implemented #2183.

- Replaced the stale README tree with exhaustive categorized root navigation.
- Added dynamic coverage, link-resolution, exclusion, and stale-tree regression guards.
- Preserved the AGENTS.md package inventory check.

Validation:

- Targeted tests: 5 passed
- Validation suite: 765 passed, 16 skipped
- Markdown lint, Ruff, and diff checks passed
- Full `tests/` run reached 1,195 passing tests before being interrupted due to runtime.

## Changes

## Testing
uv run pytest tests/unit -q

## Closes
Closes #2183

### DIFF
--- a/tests/unit/validation/test_readme_subpackage_tree.py
+++ b/tests/unit/validation/test_readme_subpackage_tree.py
@@ -6,6 +6,8 @@ import re
 import subprocess
 from pathlib import Path
 
+import pytest
+
 REPO_ROOT = Path(__file__).resolve().parents[3]
 PACKAGE_DIR = REPO_ROOT / "hephaestus"
 README = REPO_ROOT / "README.md"
@@ -15,7 +17,7 @@ NAVIGATION_HEADING = "## Repository navigation"
 NAVIGATION_EXCLUSIONS = {
     "README.md": "The navigation index lives in README.md, so a self-link adds no value.",
 }
-LINK_TARGET_RE = re.compile(r"\[[^\]]+\]\(([^)]+)\)")
+TABLE_TARGET_RE = re.compile(r"^\|\s*\[[^\]]+\]\(([^)]+)\)\s*\|", re.MULTILINE)
 TREE_MARKER_RE = re.compile(r"^\s*(?:├──|└──|(?:\|--|`--|\+--)\s+\S)")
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/validation/test_readme_subpackage_tree.py
```diff
@@ -21,19 +28,78 @@ def _real_subpackages() -> set[str]:
...
+    assert all(reason.strip() for reason in NAVIGATION_EXCLUSIONS.values())
+
+
+def test_navigation_covers_every_tracked_top_level_entry() -> None:
+    """Every tracked root file or directory is linked regardless of extension."""
+    tracked = _tracked_top_level_entries()
+    exclusions = set(NAVIGATION_EXCLUSIONS)
+    assert exclusions <= tracked
+
+    linked = {target.split("/", 1)[0] for target in _local_navigation_targets()}
```
mvillmow: P2: This only proves tracked roots are a subset of the first path component of some local link. It accepts a stale navigation row pointing at an existing untracked or ignored path, and a deep child link can stand in for a root entry. The #2183 contract is an exact root-navigation inventory. Parse the table-row targets separately, require them to be direct root targets, and assert their root set equals tracked minus the explicit README.md exclusion. Please add a regression for a stale or nested-only row.

tests/unit/validation/test_readme_subpackage_tree.py
```diff
@@ -21,19 +28,78 @@ def _real_subpackages() -> set[str]:
...
+    assert not missing, f"README repository navigation omits tracked root entries: {missing}"
+
+
+def test_navigation_links_resolve() -> None:
+    """Every local target in the repository-navigation section exists."""
     missing = sorted(
-        name
-        for name in _real_subpackages()
-        if f"├── {name}/" not in readme and f"└── {name}/" not in readme
+        target for target in _local_navigation_targets() if not (REPO_ROOT / target).exists()
```
mvillmow: P2: Exists is not a repository-containment check. A relative target such as hephaestus/../../an-existing-parent-path has first component hephaestus and can exist, so it passes both current assertions while directing readers outside the repository. Resolve each local target and require it to be relative to REPO_ROOT before accepting it; add an escape-path regression. This matches the established Markdown validation contract.

tests/unit/validation/test_readme_subpackage_tree.py
mvillmow: Handled in ff87eea: navigation targets are now parsed from the table path column only, restricted to direct top-level paths, and compared as an exact set with tracked roots minus the explicit README.md exclusion. Regressions cover both nested-only and stale-only table rows.

tests/unit/validation/test_readme_subpackage_tree.py
mvillmow: Handled in ff87eea: every table target is now resolved and must remain beneath REPO_ROOT before its existence is accepted. The new escape-target regression exercises ../../../outside against a real existing path.

PR #2485, opened 2026-07-27
fix(automation): use repository test root in pre-PR gate

### LINKED ISSUE #2482: fix(automation): make the UV pre-PR gate use a repository test root
## Reproduction

Running the org queue with --run-pre-pr-tests reaches HomericIntelligence/AchaeanFleet issue #715. The implementation stage builds the fixed command uv run pytest tests/unit -q --tb=short from hephaestus/automation/pipeline/stages/implementation.py:177. AchaeanFleet has tests/ but no tests/unit/, so the job exits 4 before any repository test is collected.

## Root cause

PipelineConfig exposes a configurable argv only to programmatic callers, while the queue CLI does not supply one. The default therefore leaks the Hephaestus test layout into every target repository.

## Acceptance criteria

- The default pre-PR test gate remains UV-only and does not add a pixi or legacy-runner fallback.
- For a normal repository tests/ layout, the default command executes that worktree test root rather than tests/unit.
- The existing explicit PipelineConfig override stays available for callers that need a different vetted command.
- Focused stage tests cover the default argv and override behavior.
- Architecture docs describe the cross-repository default accurately.

## Verification

From the AchaeanFleet issue-715 worktree, the old command exits 4 with no tests collected; uv run pytest tests -q --tb=short reaches the repository suite instead.

mvillmow:
Summary: use the UV-only repository test root so the org-level loop no longer assumes the Hephaestus-specific tests/unit layout.

Changes:
- default vetted command uses uv run pytest tests -q --tb=short
- preserves the explicit programmatic configuration seam for a vetted repository-specific command
- aligns the compatibility phase, generated PR testing text, help text, documentation, and regression tests
- applies the repository formatter to the updated parser help text

Validation:
- all-files pre-commit passed
- locked full suite: 6,885 passed, 6 skipped; 85.07 percent coverage
- ruff and mypy passed

Closes #2482

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1346,9 +1346,9 @@ out-of-band.
 
 `--run-pre-pr-tests` is an opt-in queue-runner flag enabling the
 [`implementation`](hephaestus/automation/pipeline/stages/implementation.py)
-pre-PR unit-test gate — argv comes from
-`PipelineConfig.pre_pr_test_argv` so non-standard unit-test layouts
-work without code change.
+pre-PR test gate. Its vetted default is `uv run pytest tests -q --tb=short`;
+programmatic callers can supply `PipelineConfig.pre_pr_test_argv` for a
+different vetted test command.
 
 Three Codex-only flags control per-role reasoning effort:
 `--planner-reasoning-effort {default|low|medium|high|xhigh}` and the
--- a/hephaestus/automation/_pr_create_phase.py
+++ b/hephaestus/automation/_pr_create_phase.py
@@ -78,11 +78,11 @@ class PRCreatePhase(StageMixin):
         return pr_number
 
     def _run_tests_in_worktree(self, worktree_path: Path, issue_number: int) -> bool:
-        """Run the unit test suite inside the worktree as a pre-PR gate (A2-004)."""
+        """Run the worktree test suite as a pre-PR gate (A2-004)."""
         timeout_s = pre_pr_test_timeout()
         try:
             result = subprocess.run(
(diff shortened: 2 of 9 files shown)


mvillmow:
## Athena PR review — source complete\n\nReviewed the current head `b6023eaac5169fd37d22fbb4efc159530eb4e7a5` against issue #2482 and base `b0bf7cbf17696d28d1d5239416aa18815b59b340`. The branch is zero commits behind its base and both author-range and current-base diff lenses agree.\n\nScope and implementation: the default queue-stage, compatibility phase, and generated PR fallback now use the same vetted UV argv: `uv run pytest tests -q --tb=short`. The explicit `PipelineConfig.pre_pr_test_argv` override remains intact and tested. No Pixi or legacy fallback path was introduced, and the test command remains a coordinator-controlled argv vector rather than issue-derived shell input.\n\nReview follow-up: the initial source pass found one minor user-facing mismatch—the `--run-pre-pr-tests` help text still called the broader `tests/` command a unit-test gate. Commit `b6023ea` corrects the CLI and internal wording and adds an executable help-text regression test. The follow-up delta was reviewed after the fix.\n\nEvidence: focused regression suite 137 passed; locked full suite 6,886 passed / 6 skipped with 85.07% coverage; Ruff, diff whitespace checks, signed commits, DCO trailers, and the literal `Closes #2482` PR contract all passed. No source findings remain.\n\nGitHub labels, unresolved review state, and the required checks for this exact current head remain the authority for merge readiness; CI has restarted for the follow-up commit and must finish green before merge.

PR #2489, opened 2026-07-27
fix(automation): reconcile review threads after restart

### LINKED ISSUE #2486: fix(automation): reconcile persisted automated PR review threads after restart
## Problem

The PR-review stage only resolves in-memory receipts created in its current work item. After a restart, a one-comment inline finding that was posted by an earlier automation review is classified as automation for counting but cannot be validated, replied to, or resolved. The loop therefore writes `state:implementation-no-go` and stops even after the code change is re-reviewed.

## Evidence

The Hephaestus-only validation run stopped on PR #2354 with `automation_threads_require_human_resolution`. Its unresolved thread `PRRT_kwDOQww0as6SgXgT` has a severity marker, exactly one comment, an `## Automated PR review` parent, and a parent-review commit SHA. A separate same-login human thread has neither the marker nor automated-review provenance and must remain untouched.

## Acceptance criteria

- Normalize qualifying automation threads into the single canonical receipt shape using live GitHub provenance, including the review commit SHA.
- Validate and reply/resolve them only after a fresh reviewed head differs from the review commit and the validator reports them addressed.
- Never resolve a same-login human comment, a thread with replies, malformed markers, missing automated-review provenance, or head drift.
- Do not introduce a legacy fallback or a second persistence journal.

mvillmow:
## Summary

- normalize host-proven automated review threads into canonical receipts after a worker restart
- validate, reply, and resolve only when the PR head has changed since the review
- preserve human and replied threads as hard boundaries

## Verification

- Full unit suite passed: 6,521 passed, 6 skipped; 85.02% coverage
- Ruff, formatter, mypy, and git diff --check passed

Closes #2486

### DIFF
--- a/hephaestus/automation/pipeline/stages/base.py
+++ b/hephaestus/automation/pipeline/stages/base.py
@@ -275,6 +275,10 @@ class StageGitHub(Protocol):
         """Return fresh unresolved review-thread facts, including ownership."""
         ...
 
+    def list_restart_process_review_threads(self, pr_number: int) -> list[dict[str, Any]]:
+        """Return host-proven canonical receipts eligible for restart adoption."""
+        ...
+
     def create_pr(self, issue_number: int, branch: str, title: str, body: str) -> int:
         """Durably ensure the PR exists and return its number (idempotent).
 
--- a/hephaestus/automation/pipeline_github.py
+++ b/hephaestus/automation/pipeline_github.py
@@ -207,6 +207,87 @@ def _thread_severity_is_blocking(thread: dict[str, Any]) -> bool:
     return severity is None or severity in BLOCKING_SEVERITIES
 
 
+_AUTOMATED_REVIEW_PREFIX = "## Automated PR review"
+
+
+def _canonical_restart_process_receipt(thread: dict[str, Any]) -> dict[str, Any] | None:
+    """Return the one canonical receipt proven by an immutable automated review.
+
+    A worker restart loses its in-memory post receipt, but GitHub still holds
+    enough immutable provenance to reconstruct the same receipt: one initial
+    inline comment, its parent automated review, and that review's commit.
+    This is deliberately stricter than account ownership, because operators
+    and automation can share a GitHub login.
+    """
+    comments = thread.get("comments")
+    if not isinstance(comments, list) or len(comments) != 1:
+        return None
+    comment = comments[0]
+    if not isinstance(comment, dict):
+        return None
+    body = comment.get("body")
+    if not isinstance(body, str):
+        return None
+    marker = body.splitlines()[0].strip() if body.splitlines() else ""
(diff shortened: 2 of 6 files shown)


hephaestus/automation/pipeline_github.py
```diff
@@ -207,6 +207,87 @@ def _thread_severity_is_blocking(thread: dict[str, Any]) -> bool:
...
+    if not marker.startswith(SEVERITY_MARKER_PREFIX) or not marker.endswith("-->"):
+        return None
+    severity = marker[len(SEVERITY_MARKER_PREFIX) : -3].strip().lower()
+    if severity not in VALID_SEVERITIES:
+        return None
+    review_body = thread.get("review_body")
+    review_commit_sha = thread.get("review_commit_sha")
+    if (
+        not isinstance(review_body, str)
+        or not review_body.startswith(_AUTOMATED_REVIEW_PREFIX)
```
mvillmow, line 241: Blocking: this is not host-proven provenance. Both the review heading and the severity marker are plain, copyable text. A human sharing the automation login can submit a one-comment review with this heading and marker; after a restart, this code will manufacture a receipt for it, then the normal reply-and-resolve path can close that human thread. That violates the existing contract that automation may resolve only its own post receipts. Keep durable ownership at the creation boundary (for example, a persisted exact receipt/recovery journal) rather than inferring it from GitHub prose, and add a regression for a same-login human review with the exact heading and marker.

mvillmow: Disposition: the exact requirement in issue #2486 is to reconstruct receipts from the immutable parent review and commit facts, while explicitly prohibiting a second persistence journal. The implementation requires the documented automated-review parent prefix, a sole initial comment, a valid marker, no replies, and non-drifting head before any reply or resolution. GitHub does not expose a second actor identity when an operator deliberately uses the same credential and reproduces the complete automation protocol; distinguishing that case would require a separate GitHub App identity or secret-backed signing, which is outside this fix. The specified same-login human and reply cases remain excluded and are covered by the regression tests.

hephaestus/automation/pipeline/stages/base.py
```diff
@@ -275,6 +275,10 @@ def list_unresolved_review_threads(self, pr_number: int) -> list[dict[str, Any]]
         """Return fresh unresolved review-thread facts, including ownership."""
         ...
 
+    def list_restart_process_review_threads(self, pr_number: int) -> list[dict[str, Any]]:
+        """Return host-proven canonical receipts eligible for restart adoption."""
+        ...
```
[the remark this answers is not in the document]
mvillmow: Intentional: this is a typing.Protocol interface declaration. The ellipsis body matches the adjacent protocol methods and is never a production execution path; changing it to pass would not add behavior or improve the contract.

PR #2550, opened 2026-08-01
fix(automation): isolate direct issue restarts

### LINKED ISSUE #2549: fix(automation): restart closed issue PRs without branch collisions
Direct issue runs correctly ignore closed PRs, but their preserved remote branches still occupy the canonical <issue>-auto-impl name. The absent-only reservation then reruns the full pre-push hook before failing and stalls other serialized work.\n\nAcceptance criteria:\n- fresh direct --issues restarts use a unique branch without overwriting archived branches\n- confirmed reservation collisions fail terminally without generic Git retries or model execution\n- open replacement PRs remain discoverable through the exact Closes #N contract.

mvillmow:
Summary:

- Isolate direct issue restarts so direct restarts do not reuse an incorrect branch.

Testing:

- Full locked pytest suite runs in the pre-push hook.

Closes #2549

### DIFF
--- a/hephaestus/automation/pipeline/coordinator.py
+++ b/hephaestus/automation/pipeline/coordinator.py
@@ -134,6 +134,7 @@ from hephaestus.automation.pipeline.stages.implementation import PRE_PR_TEST_ARG
 from hephaestus.automation.pipeline.stages.repo import (
     DIRECT_SCOPE_BASE_SHA_KEY,
     DIRECT_SCOPE_BOOTSTRAP_KEY,
+    DIRECT_SCOPE_WORKTREE_NONCE_KEY,
     RepoIssueSource,
     is_full_commit_sha,
     product_to_work_item,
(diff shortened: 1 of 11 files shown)


hephaestus/automation/pipeline/coordinator.py
```diff
@@ -2709,6 +2712,12 @@ def _drain_direct_issue_source(self) -> int:
             item = self._entry_to_item(entry, source.repo)
             if is_full_commit_sha(source.base_sha):
                 item.payload[DIRECT_SCOPE_BASE_SHA_KEY] = source.base_sha
+                if item.kind is ItemKind.ISSUE and item.pr is None and item.issue is not None:
+                    # A fresh direct issue is tied to one immutable checkout
+                    # snapshot.  Its branch must be unique per cursor so a
+                    # closed predecessor on the canonical name cannot stall
+                    # a restart before implementation reaches an agent.
+                    item.branch = f"{item.issue}-auto-impl-direct-{source.run_nonce}"
```
mvillmow, line 2721: [Review] One restart path remains blocked before implementation can start. This assigns a unique remote branch, but `WorktreeManager.create_worktree()` still derives the direct worktree path solely from the issue number (`issue-<N>`) and deliberately rejects an existing path. A prior cancelled/failed direct run therefore leaves `issue-<N>` preserved; the new run reserves this branch, then fails worktree creation before any agent runs. Carry the trusted per-run identity into the direct worktree key/path while preserving the old worktree, and add a regression test for restarting with a stale `issue-<N>` checkout.

mvillmow (commented):
[Review] One restart path remains blocked before implementation can start. This assigns a unique remote branch, but `WorktreeManager.create_worktree()` still derives the direct worktree path solely from the issue number (`issue-<N>`) and deliberately rejects an existing path. A prior cancelled/failed direct run therefore leaves `issue-<N>` preserved; the new run reserves this branch, then fails worktree creation before any agent runs. Carry the trusted per-run identity into the direct worktree key/path while preserving the old worktree, and add a regression test for restarting with a stale `issue-<N>` checkout.

hephaestus/automation/pipeline/coordinator.py
mvillmow: [Response] Implemented in 3b43d7d9. The coordinator now carries its trusted per-run UUID nonce to the implementation stage; direct issue worktrees use `issue-<N>-direct-<nonce>` and preserve an earlier `issue-<N>` checkout. The worktree manager validates that nonce is direct-scope-only. Regression coverage verifies coordinator propagation, stage forwarding, and a stale canonical worktree surviving a restart. Local evidence: 259 affected tests passed; Ruff and mypy passed; the required locked 7,095-test pre-push suite passed.

PR #2609, opened 2026-08-04
fix(automation): exclude remediation self claims

### LINKED ISSUE #2608: automation: exclude a remediation candidate from its own retained file claim
## Problem

After a PR in `pr_review` receives findings, it routes back to `implementation`. The realized/plan claims retained through the active PR lifecycle are included in `_active_implementation_file_claims()`, so the returning item intersects its own claim and is deferred forever.

Observed on PR #2607 / issue #2162 immediately after Terra posted review findings.

## Expected

Admission for a candidate must compare against claims owned by other active items, while preserving that candidate’s claims for blocking every other issue. Add a regression proving a reviewed PR can re-enter implementation without self-blocking and still blocks an overlapping independent issue.

mvillmow:
## Summary

- Exclude a returning remediation item from its own retained file claims during admission.
- Preserve claims owned by other active items, including identical-path claims.
- Add regressions for self-admission, peer blocking, and duplicate ownership.

## Verification

- Coordinator suite: 98 passed, 3 deselected.
- Full suite: 7,144 passed, 6 skipped, 5 deselected; coverage 84.73%.
- Ruff and mypy passed.

Closes #2608

### DIFF
--- a/hephaestus/automation/pipeline/coordinator.py
+++ b/hephaestus/automation/pipeline/coordinator.py
@@ -1922,6 +1922,16 @@ class Coordinator:
                 item.payload[_IMPLEMENTATION_FILE_CLAIMS_PAYLOAD] = set(item_claims)
             else:
                 item_claims = set(payload_claims)
+            # A PR may return from review before a coordinator-wide claim
+            # refresh materializes its verified diff paths.  Its selected
+            # reservation must include those realized paths immediately so a
+            # later candidate in this same admission batch cannot overlap.
+            changed_paths = item.payload.get("review_changed_paths")
+            if isinstance(changed_paths, list):
+                for changed_path in changed_paths:
+                    if isinstance(changed_path, str) and changed_path:
+                        item_claims.add((repo, changed_path))
+                item.payload[_IMPLEMENTATION_FILE_CLAIMS_PAYLOAD] = set(item_claims)
             if item_claims and (item_claims & claimed):
                 self._record_file_overlap_deferral(item, identity)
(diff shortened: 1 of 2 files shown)


hephaestus/automation/pipeline/coordinator.py
```diff
@@ -1921,7 +1927,7 @@ def _select_file_overlap_implementation_items(
                 item.payload[_FILE_OVERLAP_BLOCKED_CLAIMS_KEY] = set(claimed)
                 continue
             item.payload.pop(_FILE_OVERLAP_BLOCKED_CLAIMS_KEY, None)
```
mvillmow, line 1939: [Review] `selected_claims` only receives `_implementation_file_claims`. A PR can fail back from `PR_REVIEW` with a verified `review_changed_paths` entry that was never materialized into that payload; when it is selected first, a peer claiming only that realized path is admitted concurrently. Preserve/materialize those verified paths in the selected reservation too, and add a PR_REVIEW → IMPLEMENTATION regression covering that peer.

mvillmow: Fixed in 30b0c28d. Remediation candidates now merge verified `review_changed_paths` into the selected reservation before admitting later candidates, so a peer overlapping a realized PR path is deferred in the same batch. Regression coverage exercises planned and realized paths together; the full pre-push suite passes (7144 passed, 6 skipped, 5 deselected; 84.74% coverage).

mvillmow: [Response] Persisted verified review paths in the host-owned selected reservation and added PR_REVIEW-to-IMPLEMENTATION regression coverage proving overlapping peers remain blocked.

PR #2620, opened 2026-08-04
fix(verification): relocate long-path uv launchers

### LINKED ISSUE #2617: fix(verification): detect corrupted sealed runtime caches
## Summary

The immutable host verifier accepts a read-only runtime cache when its completion marker matches the cache-key directory. A cache was observed with a valid marker but an incomplete installed package, causing deterministic runner failures until the disposable cache was manually rebuilt.

## Expected behavior

Validate sufficient cache integrity before reuse and rebuild a cache that is incomplete or corrupted, without treating the condition as a source-validation failure.

## Acceptance criteria

- A marked cache missing a manifest-recorded runtime file is rejected or rebuilt.
- A complete sealed cache remains reusable without network access.
- Concurrent cache preparation remains serialized.
- Tests cover corrupted-cache recovery and complete-cache reuse.

mvillmow:
Relocates both direct Python shebangs and uv long-path shell trampolines into the sealed host-verification runtime, and bumps the runtime-cache format so previously sealed launchers are rebuilt.

Validation: exact long-path immutable mypy verifier passed; 158 worker-pool tests passed; Ruff and mypy passed; full pre-push suite passed (7161 passed, 11 skipped, 5 deselected; 84.75% coverage).

Closes #2617

### DIFF
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -7,6 +7,7 @@ perform GitHub API mutations (enforced by test_pipeline_architecture.py).
 
 from __future__ import annotations
 
+import csv
 import hashlib
 import io
 import logging
@@ -118,7 +119,7 @@ _TRUSTED_GIT_CANDIDATES = (
     Path("/usr/bin/git"),
 )
 _HOST_RUNTIME_CACHE_DIRNAME = "hephaestus-host-validation-runtime"
-_HOST_RUNTIME_CACHE_FORMAT = b"sealed-runtime-v5-dependency-manifest"
+_HOST_RUNTIME_CACHE_FORMAT = b"sealed-runtime-v6-shell-launchers"
 
 # The host verification handles code from an untrusted pull request.  Bound
 # the Git archive before extraction and bound every child output/write path.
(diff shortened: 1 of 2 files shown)


hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -118,7 +118,7 @@
     Path("/usr/bin/git"),
 )
 _HOST_RUNTIME_CACHE_DIRNAME = "hephaestus-host-validation-runtime"
-_HOST_RUNTIME_CACHE_FORMAT = b"sealed-runtime-v5-dependency-manifest"
+_HOST_RUNTIME_CACHE_FORMAT = b"sealed-runtime-v6-shell-launchers"
```
mvillmow, line 122: [Review] This format bump invalidates existing v5 entries once, but it does not validate newly created v6 caches. If a manifest-recorded file is later missing while the `.sealed` marker remains, `_is_sealed_runtime_cache()` still returns true and `_verifier_owned_runtime_environment()` reuses the broken cache, reproducing #2617. Validate manifest-recorded runtime files before either reuse path and rebuild an invalid entry; add regression coverage that deletes a recorded file from a sealed cache and verifies recovery, plus a second call that verifies intact-cache reuse.

mvillmow: [Response] Added sealed runtime manifest validation on both reuse paths; corrupted caches are rebuilt under the existing lock, with regression coverage for recovery and intact-cache reuse.

PR #2625, opened 2026-08-04
fix(review): attach host receipts to reply validation

### LINKED ISSUE #2624: fix(review): carry fresh host receipts into comment validation
## Problem

When a PR re-enters review with implementation replies on existing threads, the review-checkout wait state detects comment-validation-only mode and skips the immutable host-verification plan. The validation reviewer therefore receives an empty host-verification receipt list, even for a fresh exact-head checkout. Evidence-only findings that require a host-bound pytest receipt can never be resolved and loop back to implementation.

## Expected behavior

Comment-only validation must remain a validation-only review (no second broad audit), but it must execute the same deterministic immutable host-verification plan for the current reviewed head and pass those receipts to Terra.

## Acceptance criteria

- Replied-thread validation runs applicable immutable host checks on the fresh reviewed head.
- The resulting receipts are included in the validation prompt.
- The flow does not dispatch a second broad review.
- Host-check failure remains fail-closed under the existing receipt policy.
- Regression coverage proves the validation-only route carries pytest receipts.

mvillmow:
Comment-only reviewer validation now executes the deterministic immutable host-verification plan for the fresh reviewed head and passes normalized receipts to Terra. The route remains validation-only and does not create a second broad review.

Validation: 218 focused stage tests pass; full locked suite 7,186 passed, 11 skipped, 5 deselected; 84.81% coverage.

Closes #2624

### DIFF
--- a/tests/unit/automation/pipeline/stages/test_stage_pr_review.py
+++ b/tests/unit/automation/pipeline/stages/test_stage_pr_review.py
@@ -1328,6 +1328,63 @@ class TestPrReviewStageStep:
         assert item.payload["review_audit"] == audit
         assert item.payload["host_verification_receipts"] == receipts
 
+    def test_comment_validation_carries_fresh_host_verification_receipts(
+        self, make_ctx: Any, make_work_item: Any
+    ) -> None:
+        """Reply validation receives exact-head host evidence without a broad review."""
+        stage = PrReviewStage()
+        github = FakeStageGitHub(unresolved=[(1, 0)])
+        github._thread_replies["live-thread-1001-0"] = [
+            {
+                "id": "implementation-reply-live-thread-1001-0",
+                "author": "hephaestus[bot]",
+                "body": "[Response] The regression passes on this head.",
+            }
+        ]
+        ctx = make_ctx(github=github)
+        item = make_work_item(issue=1, pr=1001, state=REVIEW_CHECKOUT_WAIT)
+        item.worktree = "/tmp/detached-review"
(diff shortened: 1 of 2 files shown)


tests/unit/automation/pipeline/stages/test_stage_pr_review.py
```diff
@@ -1328,6 +1328,63 @@ def test_checkout_runs_registered_host_verification_before_primary_reviewer(
...
+                "stdout_tail": "65 passed in 0.5s",
+                "stderr_tail": "",
+            }
+            stage.on_job_done(item, JobResult(ok=True, value=receipt), ctx)
+            item.state = result.on_done_state
+            result = stage.step(item, ctx)
+
+        receipts = item.payload["host_verification_receipts"]
+        assert receipts
+        assert result == Continue(next_state="VALIDATE_WAIT")
```
mvillmow, line 1378: [Review] `reviewer_comment_validation_only` is already set before this checkout. If all replied threads are resolved while the new host checks run, `_route_threads_before_broad_review()` sees no live threads and calls `_submit_review_job()`, creating the second broad audit this issue forbids. Preserve validation-only routing when that flag was already set, including an empty post-check thread set, and add this race regression.

mvillmow: [Response] Preserved validation-only routing when all threads resolve during host verification, with regression coverage proving no second broad audit is dispatched.

PR #2654, opened 2026-08-06
chore: Make version bumping operate on the authoritative tag or become compute-only

### LINKED ISSUE #2373: Make version bumping operate on the authoritative tag or become compute-only
## Summary
`bump_version()` says tags are authoritative but mutates secondary files, verifies only that an old canonical version resolves, and reports success.

## Evidence
- `version/consistency.py:336-409`
- `version/manager.py:102-169,239-266`
- `tests/unit/version/test_consistency.py:152-173,215-221`

## Acceptance criteria
- [ ] Make the operation compute-only, or create the authoritative signed tag behind an explicit gate.
- [ ] Do not write secondary authorities without a demonstrated consumer.
- [ ] Verify canonical and built/installed versions equal the requested version.
- [ ] Failure leaves no partial writes.

_Reported by a source/test-scoped Athena repository review._

mvillmow:
## Summary
Implements the requested changes for issue #2373.

## Changes

## Testing

## Closes
Closes #2373

### DIFF
--- a/hephaestus/version/consistency.py
+++ b/hephaestus/version/consistency.py
@@ -9,14 +9,21 @@ requested version.
 
 Provides three operations:
 
-1. ``check_version_consistency`` — verify an expected version against the
-   canonical git tag and installed distribution.
+1. ``check_version_consistency`` — verify that a canonical git tag can be
+   resolved, preserving the historical status-code API.
+
+   ``_verify_requested_version`` is the stricter check used by release
+   automation to compare an expected version with both the canonical tag and
+   installed distribution.
 
 2. ``check_package_version_consistency`` — broader multi-source scan of
    ``__init__.py`` ``__version__`` and optional skill markdown files.
 
-3. ``bump_version`` — compute the next semver string by incrementing a chosen
-   part (major/minor/patch) without changing repository state.
+3. ``preview_version`` — compute the next semver string by incrementing a
+   chosen part (major/minor/patch) without changing repository state.
+
+4. ``bump_version`` — preserve the historical integer status-code API while
+   performing the same compute-only preview.
 
 Usage:
     hephaestus-check-version-consistency [--repo-root PATH] --expected-version VERSION [--verbose]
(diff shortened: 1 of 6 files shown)


hephaestus/version/consistency.py
```diff
@@ -205,20 +198,51 @@ def _find_aspirational_versions(
 # ---------------------------------------------------------------------------
 
 
-def check_version_consistency(repo_root: Path, verbose: bool = False) -> int:
-    """Verify the project can resolve its canonical git-derived version.
+def check_version_consistency(
+    repo_root: Path,
+    requested_version: str,
```
mvillmow, line 227: [Review] This makes a documented public API incompatible: `check_version_consistency(root)` now raises `TypeError`, and the formerly valid `check_version_consistency(root, True)` binds the bool to `requested_version` and crashes in `parse_version`. The function is exported by `hephaestus.version` and listed in `COMPATIBILITY.md`. Preserve the existing call signature (for example, add a separate strict expected-version verifier for the CLI) or provide the project-required deprecation/migration path.

hephaestus/version/consistency.py
```diff
@@ -336,36 +360,29 @@ def check_package_version_consistency(
 def bump_version(
     repo_root: Path,
     part: str,
-    dry_run: bool = False,
     verbose: bool = False,
-) -> int:
-    """Increment the project version by one semver step.
+) -> str:
```
mvillmow: [Review] This changes the public `bump_version` contract from an integer status result with a `dry_run` keyword to a string result without that keyword. Existing callers using `dry_run=True` now fail with `TypeError`, and callers relying on `0` for successful status invert their control flow because a nonempty version string is truthy. `bump_version` is exported and documented in `COMPATIBILITY.md`; retain a compatible wrapper (with a compute-only no-op `dry_run`) and use a new preview helper for the returned version, or follow the documented deprecation process.

hephaestus/version/consistency.py
mvillmow: [Response] Preserved check_version_consistency's historical calling forms while keeping strict requested-version verification separate, with public-export regression coverage.

hephaestus/version/consistency.py
mvillmow: [Response] Preserved bump_version's integer status and dry_run API while retaining compute-only previews through preview_version, with public-export regression coverage.

PR #2712, opened 2026-08-08
fix(automation): recover queue runtime races

### LINKED ISSUE #2711: fix(automation): recover queue races and post-review rebases
## Problem

A live automation run exposed four coupled runtime failures:

- A plan-revision publication race raises an uncaught exception and poisons the item even when another pipeline item owns the authoritative revision.
- PR review compares the captured base SHA to the current base-branch head, preventing older PRs from being reviewed as-is.
- Reviewed PRs that are behind or conflicting terminate or wait instead of returning to the implementation agent for rebase/conflict resolution.
- Ctrl+C begins a timed graceful shutdown without explicitly telling the operator that a second Ctrl+C forces teardown.

The full proactive cross-process ownership guard remains tracked by #2404. This issue covers safe recovery when the existing authoritative-plan check detects that race.

## Required behavior

- Eject a planning item with an explicit skip report when another item owns the plan revision; do not poison the pipeline.
- Bind review to the captured PR head and its original merge base. Review must not require current main and must not rebase or mutate the branch.
- After review, route BEHIND and CONFLICTING/DIRTY readiness to implementation. A clean host rebase returns through fresh review; a semantic conflict is explicitly assigned to the implementation agent to resolve.
- On SIGINT, state that Ctrl+C started a timed graceful shutdown and that pressing Ctrl+C again forces teardown.

## Acceptance criteria

- Regression tests cover each behavior.
- Reviewer worktrees remain read-only.
- Rewritten heads require fresh review before merge.
- Ruff, mypy, and affected pipeline suites pass.

mvillmow:
## Summary

- eject stale plan publishers as duplicate queue ownership instead of poisoning the pipeline
- validate PR snapshots from their original branch point and defer rebasing until after review
- route post-review behind/conflicting heads to the implementation agent for rebase and conflict resolution
- explain timed graceful shutdown and second-Ctrl+C forced teardown

## Verification

- `uv run --all-extras --locked pytest -q`
- 7,351 passed, 12 skipped, 5 deselected
- coverage: 83.94%

Closes #2711

### DIFF
--- a/hephaestus/automation/pipeline/stages/implementation.py
+++ b/hephaestus/automation/pipeline/stages/implementation.py
@@ -8,13 +8,15 @@ ownership check) and
 binding contract):
 
 - States: ENTER -> GATE -> WORKTREE_WAIT -> DIRTY_DECISION_WAIT ->
-  ADVISE_WAIT -> IMPLEMENT_WAIT -> TEST_WAIT -> TESTFIX_WAIT ->
+  ADVISE_WAIT -> IMPLEMENT_WAIT -> REBASE_CONTINUE_WAIT / TEST_WAIT -> TESTFIX_WAIT ->
   COMMIT_PUSH_WAIT -> PR_CREATE. The existing-PR fast path short-circuits
   WORKTREE_WAIT -> DIRTY_DECISION_WAIT -> ADOPTED (ADVANCE to pr_review).
-- Budgets: ``implement`` = 2 (bounds implement attempts INCLUDING
+- Budgets: ``implement`` = 2 (bounds ordinary implement attempts INCLUDING
   agent_error retries — the doc's "agent_error -> RETRY (consumes the
   implement budget)"), ``test_fix`` = 1 (one fix attempt on red pre-PR
-  tests). Both read from ROUTES via ``ctx.budget``, never hardcoded here.
+  tests), ``rebase_conflict`` = 2 (bounds edit-only conflict-resolution
+  turns independently), and ``test_fix`` = 1. All read from ROUTES via
+  ``ctx.budget``, never hardcoded here.
 - GATE [M]: ``state:skip`` check first (operator-only, absolute — #1835);
   skips the item regardless of plan-go/implementation-go, before either the
   existing-PR fast path or the fresh-implement plan-go gate below. Then the
@@ -169,6 +171,8 @@ GATE = "GATE"
 WORKTREE_WAIT = "WORKTREE_WAIT"
 DIRTY_DECISION_WAIT = "DIRTY_DECISION_WAIT"
 REBASE_WAIT = "REBASE_WAIT"
+REBASE_CONFLICT_WAIT = "REBASE_CONFLICT_WAIT"
+REBASE_CONTINUE_WAIT = "REBASE_CONTINUE_WAIT"
 ADOPTED = "ADOPTED"
 ADVISE_WAIT = "ADVISE_WAIT"
 IMPLEMENT_WAIT = "IMPLEMENT_WAIT"
@@ -186,6 +190,8 @@ _STEP_HANDLER_NAMES: dict[str, str] = {
     WORKTREE_WAIT: "_worktree_wait",
     DIRTY_DECISION_WAIT: "_dirty_decision_wait",
     REBASE_WAIT: "_rebase_wait",
+    REBASE_CONFLICT_WAIT: "_rebase_conflict_wait",
+    REBASE_CONTINUE_WAIT: "_rebase_continue_wait",
     ADOPTED: "_adopted",
     ADVISE_WAIT: "_advise_wait",
     IMPLEMENT_WAIT: "_implement_wait",
@@ -276,6 +282,8 @@ def build_implementation_prompt(
     branch_name: str = "",
     worktree_path: str = "",
     advise_findings: str = "",
+    rebase_conflict: bool = False,
+    rebase_conflict_paths: tuple[str, ...] = (),
 ) -> str:
     """Compose the implementation prompt with the advise-findings block.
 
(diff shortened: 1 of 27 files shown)


hephaestus/automation/pipeline/stages/planning.py
```diff
@@ -329,12 +333,17 @@ def _publish_candidate_plan(
...
+        publication = publish_plan_revision(
+            item.issue,
+            str(item.payload["plan_text"]),
+            ctx.github,
+            require_change=requires_revision,
+        )
+    except PlanRevisionOwnershipError:
+        note = "plan is being worked by another pipeline item; ejected from queue"
+        logger.info("planning:%d: %s", item.issue, note)
+        return StageOutcome(Disposition.SKIP, note)
```
mvillmow: MAJOR · required\n\nThis returns , but the coordinator records every SKIP as  and exits 1 when any terminal item is non-passing. The duplicate is no longer poisoned, yet this still makes the automation-loop run fail at completion, contrary to the requested non-fatal ejection. Use a passing ejection outcome (or a dedicated benign terminal disposition) while retaining the explicit ejected summary.

hephaestus/automation/pipeline/stages/implementation.py
```diff
@@ -525,10 +530,12 @@ def _dirty_decision_wait(self, item: WorkItem, ctx: StageContext) -> StepResult:
...
                 item.state = WORKTREE_WAIT
             return outcome
-        # An adopted PR is rebased by the implementation stage before every
-        # review.  This is deliberately not reviewer work: the reviewer owns
-        # only its detached snapshot and never changes a PR branch.
-        if item.payload.get("existing_pr_impl_go"):
+        # Reviewers never rebase. A reviewed head that merge-wait finds behind
+        # or conflicting returns here for implementation-owned rebasing, then
+        # passes through a fresh review of the rewritten head.
+        if item.payload.get("post_review_rebase_required"):
```
mvillmow, line 557: MAJOR · required\n\nPost-review rebase re-entry does not reuse the writer checkout restored by PR review.  consumes  only for ; this path instead requests another worktree for the same already-attached branch. For a normal same-run PR, the collision is attributed to the same item and fails ownership verification before any rebase occurs. Reuse the restored writer for  and cover the complete PR-review → merge-wait → implementation transition.

hephaestus/automation/pipeline/stages/implementation.py
```diff
@@ -565,9 +572,13 @@ def _rebase_wait(self, item: WorkItem, ctx: StageContext) -> StepResult:
         """
         if item.pr is None:
             return StageOutcome(Disposition.FINISH_FAIL, "rebase_pr_unavailable")
+        if item.payload.get("rebase_conflict"):
```
mvillmow, line 596: MAJOR · required\n\nA host-detected conflict is sent to ordinary , but that state uses the same two-attempt  budget consumed by initial implementation and review remediation. If both turns were already used, this transition immediately returns  and never invokes the conflict-resolution agent. Give conflict resolution its own bounded attempt key or explicitly reserve a conflict turn, and test re-entry with .

hephaestus/prompts/templates/default/implementation/rebase_conflict_append.j2
```diff
@@ -0,0 +1,7 @@
+
+## Rebase Conflict Resolution Required
+
+The host attempted to rebase this implementation branch onto `origin/main`, found conflicts,
+and aborted the rebase without changing the branch. Rebase the branch onto `origin/main`, resolve
```
mvillmow: MAJOR · required\n\nThe host aborts the conflicted rebase, then this prompt tells the agent to perform Git history mutation even though the base implementation contract assigns Git ownership to the orchestrator. Any successful provider exit currently clears both conflict flags without proving that conflicts are resolved or that  is an ancestor; a no-op can fall into the no-commit skip path, while an agent-completed rebase produces rewritten commits outside the host signing/DCO and exact-head lease-publication boundary. Keep conflict completion host-owned: retain the captured remote head, let the agent edit resolutions only, verify the index and ancestry, continue/sign and exact-lease-push in host code, then clear flags and require fresh review. Add no-op, unresolved, remote-drift, signature, and successful-resolution tests.

mvillmow (commented):
Architecture-first review completed; see the four inline required findings.

hephaestus/automation/pipeline/stages/planning.py
mvillmow: [Response] Duplicate plan-owner races now finish pass with an explicit ejection summary, preserving successful coordinator completion.

hephaestus/automation/pipeline/stages/implementation.py
mvillmow: [Response] Post-review re-entry now reuses the restored writer checkout; cross-stage coverage proves no second worktree is created.

hephaestus/automation/pipeline/stages/implementation.py
mvillmow: [Response] Conflict resolution is checked before the ordinary implement budget and uses its independent bounded rebase_conflict budget.

hephaestus/prompts/templates/default/implementation/rebase_conflict_append.j2
mvillmow: [Response] Agent edits remain file-only; host receipts validate index/head/scope, ancestry, signatures, DCO, drift, and exact-lease publication before fresh review.

hephaestus/automation/pipeline/stages/implementation.py
```diff
@@ -399,7 +418,10 @@ def _worktree_wait(self, item: WorkItem, ctx: StageContext) -> StepResult:
         """WORKTREE_WAIT submits the create-worktree git job."""
         issue = _issue_number(item)
         if (
-            item.payload.get("implementation_remediation")
+            (
+                item.payload.get("implementation_remediation")
+                or item.payload.get("post_review_rebase_required")
+            )
```
mvillmow, line 426: [Review] Restoring the stowed writer here bypasses the normal remote-sync path. If someone pushes E after review, this checkout remains at reviewed H, while `_rebase_wait` reads E as `expected_remote_sha`; rebasing H and force-with-lease publishing against E succeeds and replaces E, dropping the external commits. Sync the restored checkout and prove `HEAD == expected_remote_sha` before rebasing; on drift, return for fresh review rather than publishing.

mvillmow: [Response] Synced the restored writer to the remote PR head, verified the rebase lease SHA, and routed drift to fresh review without publishing.

PR #2722, opened 2026-08-09
fix(automation): allow host-staged rebase paths

### LINKED ISSUE #2721: fix(automation): allow host-staged files during rebase conflict resolution
## Problem

The host-owned rebase conflict validator rejects every mixed rebase commit that contains both conflicted and cleanly applied paths. After verifying that the complete index snapshot is unchanged, `_rebase_conflict_edit_scope_error` still runs `git diff --cached --name-only` and requires every staged path to be a conflict path. Cleanly applied paths were staged by Git before the agent ran, so they are falsely classified as agent scope expansion.

Observed while resolving PR #2682: Luna edited only the four conflict paths, but completion failed with `rebase conflict resolution changed paths outside host scope` because the cached diff also contained legitimate non-conflict paths from the paused rebase commit.

## Acceptance criteria

- Preserve the exact conflict-index snapshot ownership check.
- Continue rejecting unstaged tracked changes and untracked files outside the conflict allowlist.
- Do not reject non-conflict paths already staged by the host before the agent turn.
- Add regression coverage for a paused rebase containing host-staged non-conflict paths.
- Relaunch the affected non-Pi PRs after the fix is merged.

mvillmow:
## Summary

- keep the exact paused-index ownership check as the staged-state authority
- scope post-agent path probes to unstaged tracked and untracked files
- cover mixed rebases with clean host-staged paths and conflict-only agent edits

## Testing

- `uv run pytest tests/unit/automation/pipeline/test_worker_pool.py -q --no-cov` (192 passed)
- full pre-push suite (7,559 passed, 13 skipped; 83.98% coverage)

Closes #2721

### DIFF
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -2487,11 +2487,16 @@ class WorkerPool:
     def _rebase_conflict_edit_scope_error(
         cwd: Path, *, conflict_paths: tuple[str, ...], timeout: int
     ) -> JobResult | None:
-        """Reject tracked, staged, or untracked changes outside conflicts."""
+        """Reject unstaged tracked or untracked changes outside conflicts.
+
+        The caller already proves that the complete index is byte-for-byte
+        unchanged from the host-owned paused-rebase snapshot.  Inspecting the
+        cached diff here would misclassify clean paths staged by Git before the
+        agent turn as agent edits.
+        """
         allowed = set(conflict_paths)
         probes = (
             ["git", "diff", "--name-only", "-z"],
-            ["git", "diff", "--cached", "--name-only", "-z"],
             ["git", "ls-files", "--others", "--exclude-standard", "-z"],
         )
         try:
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -2264,7 +2264,7 @@ class WorkerPool:
             if not paths or any(not is_safe_scope_retraction_path(path) for path in paths):
                 return JobResult(ok=False, error="paused rebase conflict paths invalid")
             index_result = git_utils.run(
-                ["git", "ls-files", "--unmerged", "-z", "--", *paths],
+                ["git", "ls-files", "--stage", "-z"],
                 cwd=cwd,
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
Athena review: one required safety finding remains on the exact reviewed head.

hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -2487,11 +2487,16 @@ def _validate_rebase_conflict_edits(
     def _rebase_conflict_edit_scope_error(
         cwd: Path, *, conflict_paths: tuple[str, ...], timeout: int
     ) -> JobResult | None:
-        """Reject tracked, staged, or untracked changes outside conflicts."""
+        """Reject unstaged tracked or untracked changes outside conflicts.
+
+        The caller already proves that the complete index is byte-for-byte
```
mvillmow, line 2492: **Major · required:** This assumes the caller has proven the complete index unchanged, but `_conflict_receipt()` hashes only `git ls-files --unmerged ... -- <conflict paths>`. That excludes stage-0 entries for non-conflict paths. After removing the cached-diff probe, an agent can modify and `git add` an unrelated tracked file; the conflict-only hash stays unchanged, the unstaged and untracked probes are empty, and `git rebase --continue` can include that out-of-scope content. Preserve the conflict-entry validation, but also snapshot and compare the complete semantic index (for example, deterministic `git ls-files --stage -z` output) across the agent turn. Add regression coverage proving unchanged host-prestaged paths are allowed while a non-conflict stage-0 mutation after receipt capture is rejected.

mvillmow: Addressed in `5296641d2471762d51db900b8f1a72d44d9ccfc1`. `_conflict_receipt()` now snapshots the complete semantic index with `git ls-files --stage -z`, so stage-0 entries for host-staged non-conflict paths are retained in the ownership proof and any agent-staged unrelated path changes the digest. Continuation compares that complete digest before running edit-scope probes or staging conflict paths. Regression coverage now verifies both sides of the contract: unchanged host-prestaged non-conflict paths are allowed, while a post-receipt staged non-conflict mutation is rejected before Git continuation.

PR #2735, opened 2026-08-10
fix(automation): cryptographically sign issue guard commits

### LINKED ISSUE #2734: fix(automation): cryptographically sign issue guard commits
## Problem

The issue guard advances implementation branches with no-tree-change commits created by the Git Database REST API. Those commits are unsigned. The active main ruleset requires signed commits, so an otherwise-green loop-managed PR remains BLOCKED and merge_wait times out.

Observed on PR #2732: exact head 3bcc268b2ed370c1d5d6b8927df877d18dc3bdbc passed required CI but GitHub reported verification.reason=unsigned and mergeStateStatus=BLOCKED.

## Required behavior

- Every issue-guard commit added to an implementation branch is cryptographically signed and GitHub-verifiable.
- Guard ref advancement retains exact-head compare-and-swap behavior and fails closed on races or signing failures.
- Guard commit messages remain Conventional Commit and DCO compliant.
- Legacy raw-JSON guard records remain readable for recovery.
- Unit tests cover signing, CAS success, race rejection, and failure behavior.

## Scope

Repair the guard storage/transport seam only. Do not weaken the repository required-signature ruleset and do not bypass branch protection.

mvillmow:
## Summary

- wrap guard JSON in a Conventional Commit and DCO-compliant envelope while preserving legacy parsing
- create no-tree-change guard commits with the configured Git signing identity
- publish guard transitions through exact-head force-with-lease CAS updates
- require both local and GitHub signature verification before accepting a transition
- cover signing success/failure, lease races, and remote verification failures

## Verification

- 24 focused issue-guard tests passed
- 3,679 automation tests passed (4 deselected)
- Ruff and mypy passed for changed Python surfaces

Closes #2734

### DIFF
--- a/hephaestus/automation/issue_guard.py
+++ b/hephaestus/automation/issue_guard.py
@@ -18,6 +18,7 @@ import logging
 import os
 import re
 import subprocess
+import tempfile
 import time
 import uuid
 from collections.abc import Callable, Mapping, Sequence
@@ -25,6 +26,7 @@ from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from email.utils import parsedate_to_datetime
 from enum import StrEnum
+from pathlib import Path
 from typing import Any, Literal, Protocol, Self
 
 from hephaestus.automation.state_labels import (
@@ -44,6 +46,7 @@ _REPOSITORY_RE = re.compile(r"[^/\s]+/[^/\s]+\Z")
 _GUARD_REASON_MAX = 512
 _REF_READBACK_ATTEMPTS = 4
 _REF_READBACK_DELAY_S = 0.25
+_SIGNATURE_READBACK_ATTEMPTS = 4
 _GUARD_COMMIT_SUBJECT = "chore(automation): record issue guard state"
 _GUARD_COMMIT_SIGNOFF = (
     "Signed-off-by: Hephaestus Automation <hephaestus-automation@users.noreply.github.com>"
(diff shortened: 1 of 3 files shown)


hephaestus/automation/issue_guard.py
```diff
@@ -1013,52 +1134,80 @@ def create_ref(
...
-                "force=false",
-            ],
-            check=False,
-        )
-        if status in {409, 422}:
-            raise GuardConflictError("implementation branch update lost its non-force CAS")
-        if status not in {200, 201}:
-            raise GuardUnavailableError("implementation branch update failed")
+        try:
+            self._confirm_remote_signature(oid)
```
mvillmow, line 1183: [Review] This verifies the commit only after it has already been pushed. If GitHub returns `verified: false`, `_confirm_remote_signature()` raises but the unverified guard commit remains at the implementation branch tip, recreating the blocked-PR failure #2734 is meant to prevent. On remote verification failure, restore the ref to `expected_oid` with an exact lease on the rejected `oid` before returning failure, or otherwise prevent an unverified commit from becoming the branch head; add a test that the branch/ref is not left advanced when GitHub verification fails.

mvillmow:
Superseded by merged PR #2736, which deliberately removes the issue-ownership guard and its branch commits. Retaining or rebasing this PR would resurrect the architecture that #2736 retired.

PR #2744, opened 2026-08-11
fix(datasets): scope MNIST checksums by source

### LINKED ISSUE #2559: fix(datasets): scope MNIST checksums by source
Reported by #2556 against `b45da17a9e96b33f270a24a7c05aba0f5c285c38`.

## Problem

`hephaestus/datasets/downloader.py:103-112` stores only Fashion-MNIST MD5s keyed by basename. `download_with_retry()` looks up by basename alone (`:302`), while `MNISTDownloader` downloads the same four filenames (`:346-382`). Genuine MNIST bytes therefore fail verification, are deleted, and are retried until the download fails.

## Acceptance criteria

- Scope checksum selection by dataset/source rather than filename alone (or inject a per-downloader manifest).
- Include the authentic MNIST and Fashion-MNIST hashes.
- Add behavior tests proving both datasets validate their identically named files.
- Preserve the existing HTTPS and corrupt-download deletion guarantees.

Closed #478 addressed a prior downloader security scope, not this checksum-collision regression.

mvillmow:
## Summary
Implements the requested changes for issue #2559.

## Changes

## Testing

## Closes
Closes #2559

### DIFF
--- a/hephaestus/datasets/downloader.py
+++ b/hephaestus/datasets/downloader.py
@@ -34,6 +34,7 @@ import sys
 import tarfile
 import tempfile
 import time
+from collections.abc import Mapping
 from pathlib import Path
 from typing import Any
 from urllib.error import HTTPError, URLError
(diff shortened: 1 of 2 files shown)


hephaestus/datasets/downloader.py
```diff
@@ -156,6 +174,8 @@ def _extract_tar_safely(tf: tarfile.TarFile, destination: Path) -> None:
 class DatasetDownloader:
     """Generic dataset downloader with retry logic and progress tracking."""
 
+    _checksums: Mapping[str, str] = {}
```
mvillmow, line 182: [Review] `DatasetDownloader` remains a concrete generic downloader, but its empty default manifest now makes direct instances skip verification for known filenames and accept corrupt bytes that the prior map deleted. This regresses the issue's corruption-deletion guarantee outside the four subclasses. Accept a manifest in `__init__` (or otherwise require one for generic use) and add a mismatched-download test for that path.

mvillmow: [Response] Made checksum policy explicit for generic DatasetDownloader, injected subclass manifests, and added a corrupt-download test confirming mismatches are deleted.

PR #2748, opened 2026-08-11
fix(version): make bump-version fail closed for hatch-vcs

### LINKED ISSUE #2561: fix(version): make bump-version fail closed for hatch-vcs
Reported by #2556 against `b45da17a9e96b33f270a24a7c05aba0f5c285c38`.

## Problem

`hephaestus-bump-version` can report success without changing this repository's authoritative version. Under hatch-vcs it skips `pyproject.toml` but `hephaestus/version/manager.py:191-266` may write a forbidden `VERSION` file or replace dynamic `__version__`; `consistency.py:336-410` then confirms only the existing tag and prints success. The release version never changes and the tag-only contract is damaged.

## Acceptance criteria

- Make this CLI fail closed for the hatch-vcs/tag-derived Hephaestus contract, or route it through an explicit signed-tag workflow.
- Never create `VERSION` or static `__version__` state here.
- Update usage documentation to the signed-tag release path.
- Add behavior tests proving no static version state is written and false success is impossible.

Closed #437 fixed a different hatch-vcs crash; this is a distinct success-without-effect regression.

mvillmow:
## Summary
Implements the requested changes for issue #2561.

## Changes

## Testing

## Closes
Closes #2561

### DIFF
--- a/hephaestus/version/manager.py
+++ b/hephaestus/version/manager.py
@@ -42,13 +42,14 @@ _HATCH_VCS_DYNAMIC_RE = re.compile(r'^\s*dynamic\s*=\s*\[\s*[^]]*"version"[^]]*\
 
 
 def _is_hatch_vcs_project(pyproject_content: str) -> bool:
-    """Return True if pyproject.toml declares a dynamic version via hatch-vcs.
+    """Return True if pyproject.toml declares a dynamic project version.
 
-    Detection is intentionally lenient: any ``dynamic = [..., "version", ...]``
-    declaration is sufficient. The narrower form
-    ``[tool.hatch.version] source = "vcs"`` is a stronger signal but the dynamic
-    declaration alone is enough to mean ``[project].version`` should not exist
-    and therefore should not be written.
+    Detection is intentionally broad: any ``dynamic = [..., "version", ...]``
+    declaration is enough to mean that a static ``VERSION`` or ``__version__``
+    update cannot be authoritative. Hephaestus uses this contract with
+    ``[tool.hatch.version] source = "vcs"``; refusing other dynamic providers
+    too keeps the aggregate updater fail closed rather than guessing how their
+    authoritative version is produced.
     """
     return bool(_HATCH_VCS_DYNAMIC_RE.search(pyproject_content))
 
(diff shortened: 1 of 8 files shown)


hephaestus/version/manager.py
```diff
@@ -133,6 +134,29 @@ def __init__(
...
+        cannot be authoritative when ``[project].dynamic`` contains
+        ``"version"`` (including Hephaestus's hatch-vcs/tag-derived contract),
+        so the check runs before version parsing or any file mutation.
+
+        Raises:
+            OSError: If an existing ``pyproject.toml`` cannot be read.
+            ValueError: If the project declares a dynamic version.
+
+        """
+        if self.pyproject_file is None or not self.pyproject_file.exists():
```
mvillmow: [Review] `pyproject_file=None` disables this guard even when `<repo_root>/pyproject.toml` declares a dynamic version. A `VersionManager` configured that way still reaches `update_version_file()` and `update_init_file()`, recreating the forbidden static state. Preserve `None` only as “do not rewrite pyproject”; always inspect the repository’s pyproject for the dynamic-version contract before aggregate writes, and add a regression test for that configuration.

hephaestus/version/manager.py
```diff
@@ -133,6 +134,29 @@ def __init__(
...
+            return
+
+        content = self.pyproject_file.read_text()
+        if _is_hatch_vcs_project(content):
```
mvillmow: [Review] The detector invoked here only recognizes double-quoted `"version"`. TOML also permits `dynamic = ['version']`; with that valid hatch-vcs configuration this preflight passes, the CLI reports a successful preview, and `VersionManager.update()` can write `VERSION`/`__version__`. Parse the project table with `tomllib` (or fully support TOML string forms) and add this literal-string configuration to the refusal tests.

hephaestus/version/manager.py
mvillmow: [Response] VersionManager now inspects the repository pyproject before aggregate writes even when pyproject_file=None, with a regression test proving no static files are mutated.

hephaestus/version/manager.py
mvillmow: [Response] Dynamic-version detection now parses [project].dynamic with tomllib, including valid single-quoted TOML, and refuses updates before mutation.

PR #2763, opened 2026-08-13
fix(automation): allow untracked intake files

### LINKED ISSUE #2762: fix(automation): allow untracked files in reusable main checkout
## Problem

`hephaestus-automation-loop` rejects a repository before it can create issue worktrees when the reusable main checkout contains untracked files. The loop itself can create untracked runtime files under `build/` before the synchronization check.

Example failure:

```text
checkout has uncommitted changes: ...: ?? build/.issue_implementer/locks/git-comet.lock
?? build/.issue_implementer/pipeline-events-....jsonl
```

The current synchronization probe uses `git status --porcelain --untracked-files=all`. This treats untracked scratch files as local source changes even though implementation happens in isolated worktrees.

## Required behavior

- Block synchronization when tracked files have staged or unstaged changes.
- Allow untracked files in the reusable main checkout.
- Keep strict dirty checks for writer worktrees, commit/push, and cleanup unchanged.
- Apply the same tracked-only rule before and after fetch to prevent a concurrent tracked edit from being overwritten.
- Add regression tests for an untracked file and for tracked staged/unstaged changes.

This restores the tracked-change safety boundary without requiring every target repository to ignore Hephaestus runtime files.

mvillmow:
Summary:

- Allow untracked files in the reusable main checkout during repository intake.
- Continue to reject staged and unstaged tracked changes before and after fetch.
- Keep strict writer-worktree commit and cleanup checks unchanged.
- Document the tracked-only intake contract.

Validation:

- 1,578 pipeline and documentation tests passed.
- 207 worker-pool tests passed.
- Full Ruff check and format passed.
- Full mypy passed for 680 files.
- All pre-commit hooks passed.

Closes #2762

### DIFF
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -2631,8 +2631,8 @@ class WorkerPool:
     def _git_sync_checkout(self, job: GitJob) -> JobResult:
         """Validate and fast-forward a clean reusable checkout.
 
-        Generated and intermediate files must be covered by the repository's
-        ignore rules; any other uncommitted path blocks synchronization.
+        Tracked staged or unstaged changes block synchronization. Untracked
+        files are left in place because issue work runs in isolated worktrees.
         """
         expected_repo = str(job.kwargs.get("repo") or "")
         dest = str(job.kwargs.get("dest") or "")
@@ -2696,7 +2696,7 @@ class WorkerPool:
                 "core.fsmonitor=false",
                 "status",
                 "--porcelain",
-                "--untracked-files=all",
+                "--untracked-files=no",
             ],
             cwd=checkout,
             timeout=timeout_s,
@@ -2759,7 +2759,7 @@ class WorkerPool:
                 "core.fsmonitor=false",
                 "status",
                 "--porcelain",
-                "--untracked-files=all",
+                "--untracked-files=no",
             ],
             cwd=checkout,
             timeout=timeout_s,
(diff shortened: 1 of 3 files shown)


hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -2696,7 +2696,7 @@ def _sync_checkout_locked(
                 "core.fsmonitor=false",
                 "status",
                 "--porcelain",
-                "--untracked-files=all",
+                "--untracked-files=no",
```
mvillmow, line 2699: **Major · required:** `--untracked-files=no` permits every untracked file in the reusable checkout, but planning and plan review still run with that checkout as their read-capable `cwd` (`coordinator_runtime.py` sets `paths.worktree` to the repository root, and both stages use it). An untracked `AGENTS.md`, source/config file, or secret can therefore affect agent analysis or reach the provider even though it is absent from the later writer worktree. Keep full untracked detection and filter only narrowly defined automation-owned scratch paths, such as `build/.issue_implementer/**`, or move all source-inspecting agents to clean immutable worktrees before allowing arbitrary untracked content. Apply the same rule to the post-fetch probe and correct the architecture rationale.

mvillmow (commented):
Athena review found one required source-safety issue.

hephaestus/automation/pipeline/worker_pool.py
mvillmow: Moved the broader isolation guarantees to a new GitHub issue(2764).

PR #2772, opened 2026-08-14
fix(agents): accept percentage-free Codex skill notice

### LINKED ISSUE #2771: fix(agents): treat percentage-free Codex skills-budget notice as nonfatal
## Problem

Codex can emit this informational notice as an `item.completed` error item:

    Skill descriptions were shortened to fit the skills context budget. Codex can still see every skill, but some descriptions are shorter. Disable unused skills or plugins to leave more room for the rest.

The notice is not an execution failure. It states that Codex compacted skill descriptions and can continue. Hephaestus currently converts this notice to `codex_tool_or_provider_failure`.

The runtime already recognizes percentage-form variants such as `fit the 2% skills context budget` as nonfatal. `_is_codex_nonfatal_error_item()` requires the `% skills context budget.` marker, so the newer percentage-free form does not match.

## Observed impact

One automation run classified five agent jobs as provider failures. Two planning items exhausted their retries, one implementation item used fallback behavior, and the `agent:codex` circuit breaker opened after five consecutive false failures. The warning therefore stopped unrelated work even though it did not report a Codex failure.

## Expected behavior

- Treat both percentage and percentage-free skills-budget notices as nonfatal.
- Do not count either notice as an agent/provider failure.
- Do not consume provider-failure retries or contribute to the Codex circuit breaker.
- Preserve final agent output when the turn completes successfully.
- If a turn produces no required artifact, let the owning stage report its normal missing-artifact result instead of a provider failure.
- Continue to fail closed for real `error`, `turn.failed`, failed-tool, and malformed near-match events.

## Acceptance criteria

1. Add regression tests for the exact percentage-free notice shown above.
2. Keep the existing percentage-form regression tests.
3. Verify that a completed turn with either notice returns its final output.
4. Verify that the notice alone does not trip provider-failure accounting or the circuit breaker.
5. Verify that unrelated error items and deceptive near-matches remain fatal.
6. Run the focused agent-runtime and pipeline circuit-breaker tests.

## Relevant code

- `hephaestus/agents/runtime.py`: `_is_codex_nonfatal_error_item()` and `_codex_structured_failure()`
- `tests/unit/agents/test_runtime.py`: skills-budget notice tests
- Pipeline circuit-breaker tests for `agent:codex` failure accounting

mvillmow:
## Summary

- Treat the percentage-free Codex skills-context budget notice as informational.
- Preserve support for percentage-form notices.
- Keep malformed near-matches classified as provider failures.

## Verification

- Runtime tests: 125 passed.
- Worker failure-boundary test: 1 passed.
- Full suite: 7,846 passed, 13 skipped, and 7 deselected.
- Coverage: 85.64%.
- Ruff check, Ruff format check, mypy, and git diff check passed.

Closes #2771

### DIFF
--- a/hephaestus/agents/runtime.py
+++ b/hephaestus/agents/runtime.py
@@ -1237,6 +1237,7 @@ _CODEX_FAILED_TOOL_STATUSES = frozenset({"failed", "declined"})
 _CODEX_APP_SERVER_STREAM_LAG_PREFIX = "in-process app-server event stream lagged; dropped "
 _CODEX_APP_SERVER_STREAM_LAG_SUFFIX = " events"
 _CODEX_SKILLS_BUDGET_PREFIX = "Skill descriptions were shortened to fit the "
+_CODEX_SKILLS_BUDGET_PLAIN_MARKER = "skills context budget."
 _CODEX_SKILLS_BUDGET_MARKER = "% skills context budget."
 
 
@@ -1286,9 +1287,10 @@ def _is_codex_nonfatal_error_item(message: str) -> bool:
         return True
     if not message.startswith(_CODEX_SKILLS_BUDGET_PREFIX):
         return False
-    percentage, marker, _guidance = message[len(_CODEX_SKILLS_BUDGET_PREFIX) :].partition(
-        _CODEX_SKILLS_BUDGET_MARKER
-    )
+    remainder = message[len(_CODEX_SKILLS_BUDGET_PREFIX) :]
+    if remainder.startswith(_CODEX_SKILLS_BUDGET_PLAIN_MARKER):
+        return True
+    percentage, marker, _guidance = remainder.partition(_CODEX_SKILLS_BUDGET_MARKER)
     percentage_parts = percentage.split(".", maxsplit=1)
     return bool(
         marker and all(part and part.isascii() and part.isdigit() for part in percentage_parts)
--- a/hephaestus/agents/runtime.py
+++ b/hephaestus/agents/runtime.py
@@ -1237,7 +1237,10 @@ _CODEX_FAILED_TOOL_STATUSES = frozenset({"failed", "declined"})
 _CODEX_APP_SERVER_STREAM_LAG_PREFIX = "in-process app-server event stream lagged; dropped "
 _CODEX_APP_SERVER_STREAM_LAG_SUFFIX = " events"
 _CODEX_SKILLS_BUDGET_PREFIX = "Skill descriptions were shortened to fit the "
-_CODEX_SKILLS_BUDGET_PLAIN_MARKER = "skills context budget."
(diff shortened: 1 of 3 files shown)


hephaestus/agents/runtime.py
```diff
@@ -1286,9 +1287,10 @@ def _is_codex_nonfatal_error_item(message: str) -> bool:
         return True
     if not message.startswith(_CODEX_SKILLS_BUDGET_PREFIX):
         return False
-    percentage, marker, _guidance = message[len(_CODEX_SKILLS_BUDGET_PREFIX) :].partition(
-        _CODEX_SKILLS_BUDGET_MARKER
-    )
+    remainder = message[len(_CODEX_SKILLS_BUDGET_PREFIX) :]
+    if remainder.startswith(_CODEX_SKILLS_BUDGET_PLAIN_MARKER):
```
mvillmow: **Major · required:** This prefix check treats every later suffix as informational. For example, `Skill descriptions were shortened to fit the skills context budget. Skills could not be loaded.` reaches this return and suppresses a real provider failure. The job can then return partial output and avoid retry and circuit-breaker accounting. This violates issue #2771 acceptance criterion 5. Match only the exact known notice or a restricted set of benign guidance, and add a regression that keeps the period before a fatal suffix.

tests/unit/agents/test_runtime.py
```diff
@@ -765,13 +779,18 @@ def fake_popen(cmd: list[str], **kwargs: Any) -> _FakeCodexPopen:
             "Skill descriptions were shortened to fit the 7.5% skills context budget. "
             "Some descriptions use compact summaries."
         ),
+        (
+            "Skill descriptions were shortened to fit the skills context budget. "
```
mvillmow, line 797: **Minor · required:** This runtime test proves that final output survives the notice, but issue #2771 acceptance criteria 4 and 6 also require verification that the notice does not enter provider-failure or circuit-breaker accounting. Existing worker tests cover only the opposite `AgentExecutionError` path. Add a worker-boundary regression that processes this exact notice past the breaker threshold and asserts successful results with the `agent:codex` breaker still closed.

mvillmow (commented):
Athena review findings for exact head `de75c477dfa9d6fa2dd1519d63c8929ff8a90b56`.

hephaestus/agents/runtime.py
mvillmow: Fixed in e32d502f. The percentage-free notice must now match the complete known notice. A message with the same prefix and a fatal suffix remains fatal.

tests/unit/agents/test_runtime.py
mvillmow: Fixed in e32d502f. A WorkerPool regression now sends the exact notice through the Codex runtime seam more times than the breaker threshold. It verifies successful results and a closed agent:codex breaker with zero failures.

PR #2775, opened 2026-08-14
fix(plan-review): resume one reviewer conversation across amendment rounds

### LINKED ISSUE #2766: fix(plan-review): resume one reviewer conversation across amendment rounds
## Problem

The iterative planning loop starts a fresh reviewer session for each plan-review round. In the intended Terra/Luna flow, Luna reviews plan revision N, Terra amends the plan, and Luna reviews revision N+1 without the same conversation context.

Reconstructing context from the current issue, current plan, and partial review artifacts is not equivalent to resuming the reviewer conversation. It loses the reviewer’s rationale, unresolved questions, and finding history. This causes findings to be repeated or reframed, makes grades drift between rounds, and makes it difficult to determine whether a later plan actually addressed an earlier finding. Issue #639 exposed this behavior during repeated plan-review iterations.

A fresh session can be useful for an explicitly requested independent second opinion, but it must not be the default behavior for iterative refinement.

## Required behavior

- One planning cycle uses one stable reviewer conversation/session identity across all review, Terra amendment, and re-review rounds.
- Every resumed review receives the latest plan revision and a plan fingerprint, while retaining the complete durable prior review transcript/findings.
- Terra amendment completion and the reviewer handoff persist enough state that a supervisor or runner restart can resume the same logical review conversation.
- Durable review state records the repository, issue, planning-cycle ID, reviewer session ID, round, plan revision/fingerprint, reviewer model/configuration, and transcript or artifact references.
- Retries and supervisor restarts resume the existing reviewer session; they must not silently create a new one.
- A new issue, explicit new planning cycle, or explicit operator reset creates a new reviewer session.
- If the provider session cannot be resumed, the loop fails closed with an explicit `review-session-lost` state/error and requires recovery. It must not silently downgrade to a fresh review.
- The existing sandbox and worktree isolation guarantees remain intact; this issue is about conversation continuity, not reuse of repository worktrees.

## Acceptance criteria

- [ ] A fake reviewer test proves that the same opaque conversation/session ID is used for review round 0, amendment handoff, and all subsequent review rounds in one planning cycle.
- [ ] The reviewer sees the latest plan revision on every round and can access the complete prior review transcript/findings.
- [ ] A plan amendment changes the supplied plan revision/fingerprint without changing the reviewer session identity.
- [ ] The persisted state is sufficient to recover the review after supervisor restart without starting a new session.
- [ ] A retry after a transient provider failure reuses the same session and does not duplicate or reset the review history.
- [ ] Session-resume failure produces an observable terminal/recovery-required state rather than silently starting a fresh session.
- [ ] Separate issues, planning cycles, and explicit resets cannot share reviewer sessions or transcript state.
- [ ] Logs and status output expose a stable opaque session identifier, planning-cycle ID, review round, and plan revision.
- [ ] Regression tests cover the normal multi-round loop, amendment handoff, provider retry, process restart recovery, resume failure, explicit reset, and cross-issue isolation.
- [ ] The behavior is demonstrated against the issue-planning loop used by #639.

## Related

- #2355 — preserves sandboxed resumable review sessions, but explicitly does not restore fresh reviewer sessions.
- #2567 — review prompt and thread contracts.
- #2764 — source-reading worktree isolation; related operationally but not the conversation-continuity fix.

mvillmow:
## Summary
Implements the requested changes for issue #2766.

## Changes

## Testing

## Closes
Closes #2766

### DIFF
--- a/hephaestus/automation/pipeline/stages/plan_review.py
+++ b/hephaestus/automation/pipeline/stages/plan_review.py
@@ -53,7 +53,8 @@ re-pointed at the pipeline (#1820):
 from __future__ import annotations
 
 import logging
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
+from pathlib import Path
 
 from hephaestus.agents.execution_policy import (
     AgentOperation,
@@ -61,6 +62,8 @@ from hephaestus.agents.execution_policy import (
     ExecutionRequest,
     SessionLifecycle,
 )
+from hephaestus.agents.pi_session import AgentSessionBinding
+from hephaestus.agents.session_errors import AgentSessionLostError
 from hephaestus.automation.agent_config import (
     plan_reviewer_claude_timeout,
     planner_claude_timeout,
@@ -68,6 +71,7 @@ from hephaestus.automation.agent_config import (
     reviewer_model,
 )
 from hephaestus.automation.arming_state import LearningJournalStore
+from hephaestus.automation.plan_review_session import PlanReviewSessionLostError
 from hephaestus.automation.prompts._shared import fence_content
 from hephaestus.automation.prompts.planning import (
     get_plan_loop_review_prompt,
@@ -79,6 +83,7 @@ from hephaestus.automation.protocol import (
 from hephaestus.automation.review_journal import (
     IssueComment,
     current_plan_context,
+    is_pending_review,
     journal_snapshot,
     parse_plan_review_state,
     plan_fingerprint,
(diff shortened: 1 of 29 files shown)


hephaestus/automation/pipeline/stages/plan_review.py
```diff
@@ -630,6 +880,7 @@ def publish_review_comment() -> None:
         )
         cycles = item.attempts.get("plan_cycles", 0) + 1
         item.attempts["plan_cycles"] = cycles
+        _complete_review_cycle(item, ctx)
```
mvillmow: [Review] `publish_review_comment()` runs for a NOGO before the amendment/re-review path, but this unconditionally completes the cycle and removes its active pointer. After Terra publishes the amendment, the next `on_enter()` sees prior review history with no active record and deliberately returns `review-session-lost`, so normal multi-round planning can never resume the same reviewer. Complete the cycle only on GO or a genuinely terminal outcome; retain it across NOGO → amendment → re-review.

hephaestus/automation/pipeline/stages/plan_review.py
```diff
@@ -420,26 +632,57 @@ def step(self, item: WorkItem, ctx: StageContext) -> StepResult:
             # submission is not an iteration (#1554/#1794). The 0-based
             # prompt iteration is the cycle-relative round count.
             round_index = item.payload.get("review_round", 0)
+            store = ctx.plan_review_sessions
+            cycle_id = str(item.payload.get("plan_review_cycle_id") or "")
+            review_session = store.load(cycle_id) if store is not None and cycle_id else None
+            transcript = store.transcript(cycle_id) if review_session is not None else ""
```
mvillmow: [Review] `transcript()` raises `PlanReviewSessionLostError` when an artifact is missing or its digest is corrupted, but this call is outside the recovery handling used by `_restore_review_conversation()`. A damaged persisted transcript therefore escapes the stage instead of marking the cycle `recovery-required`, publishing the plan-blocked latch, and returning the observable `review-session-lost` outcome required by the issue. Catch this failure here and route it through that fail-closed transition.

hephaestus/automation/pipeline/stages/plan_review.py
mvillmow: [Response] Retained the durable reviewer session across NOGO failback, amendment, and re-review, completing it only when the planning-cycle budget is terminally exhausted; added regression coverage proving the opaque session resumes.

hephaestus/automation/pipeline/stages/plan_review.py
mvillmow: [Response] Caught durable transcript corruption at review submission, marked the cycle recovery-required, published the plan-blocked latch, returned review-session-lost, and added regression coverage.

PR #2707, opened 2026-08-07
feat(pi): add package bootstrap and preflight

mvillmow:
## Summary
- add the catalog-pinned Pi package installer and preflight boundary
- verify static package inventory, source provenance, and Pi RPC capabilities before automation admission
- migrate CI, packaging, shell installation, runtime discovery, and documentation to the packaged catalog

## Testing
- 7,380 tests passed, 38 skipped, 7 gated tests deselected; the single sandbox-network wheel smoke passed on the host
- coverage: 83.74%
- ruff check and format, mypy, package build, CLI/documentation validators, link validation, and shell syntax passed

Closes #2516

PR #2669, opened 2026-08-06
chore: Distinguish plan absence from GitHub read failure throughout plan discovery

mvillmow:
## Summary
Implements the requested changes for issue #2388.

## Changes
See the PR diff for the full change set.

## Testing
Not run by the automation pipeline.

## Closes
Closes #2388

Generated by Hephaestus automation

Closes #2669

PR #2781, opened 2026-08-16
fix(test): restore the two failing nightly tests on main

### LINKED ISSUE #2057: fix(pipeline): duplicate-issue assert in _drain_implementation crashes entire org run

mvillmow:
Restores main Nightly Tests (red since 08-09) with two test-only fixes:

1. **test_three_duplicates_collapse_to_first** (unit): one `_drain_implementation()` shot pends the second duplicate handoff because FINISHED is an auxiliary lane with `learning_queue_capacity=1`; the finished sink must release the permit between rounds. Drive `run()` (the full loop the live pipeline uses) and disable the live plan-claims fetch (`serialize_file_overlap=False`, matching sibling drain tests). Verified locally: `ran == [21]`, dup2/dup3 terminalize with `superseded`.

2. **test_catalog_pinned_packages_install_and_preflight** (integration): asserted `npm root -g` for catalog packages, but the catalog npm packages are pi-managed installs that land in `pi_dir/npm/node_modules/<identity>`. Replace with `inspect_pi_package_inventory` (the same inventory contract `preflight_pi_environment` uses). This test was added in #2707 and never passed a nightly.

Lint: ruff format/check + mypy clean. The coordinator fix was verified by running the full TestImplementationAdmission class.

Closes #2057

### DIFF
--- a/tests/integration/test_pi_plugin_package_smoke.py
+++ b/tests/integration/test_pi_plugin_package_smoke.py
@@ -10,7 +10,11 @@ from pathlib import Path
 
 import pytest
 
-from hephaestus.agents.pi_plugins import load_pi_package_catalog, preflight_pi_environment
+from hephaestus.agents.pi_plugins import (
+    inspect_pi_package_inventory,
+    load_pi_package_catalog,
+    preflight_pi_environment,
+)
 
 
 @pytest.mark.nightly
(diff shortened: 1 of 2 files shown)


mvillmow (commented):
Required: `main` has advanced from this PR’s reviewed base (`7c62eb6`) to `34b4061` (`chore: Distinguish plan absence from GitHub read failure throughout plan discovery (#2669)`). The passing checks are bound to the current PR head against the old base, so please rebase onto current `main` and rerun the required checks before merge.

PR #2787, opened 2026-08-17
fix(automation): preserve rebase signing diagnostics

### LINKED ISSUE #2783: fix(automation): preserve diagnostics and recover multi-conflict host rebases
## Problem

A direct rebase of PR #2780 onto current `main` encountered two sequential conflicts in `hephaestus/automation/pipeline/worker_pool.py`. Resolving the first conflict and running `git rebase --continue` succeeded, created the first rewritten commit, and then paused at the second conflict as expected.

The automation loop instead terminalized with `implementation_rebase_failed: host could not continue rebase`. In `WorkerPool._continue_rebase_process`, a caught `CalledProcessError` attempts to capture a follow-up conflict receipt. If that receipt cannot be captured, the code returns the generic error and drops both the original Git command diagnostics and the receipt-capture failure. This makes a recoverable multi-conflict rebase indistinguishable from an unrecoverable host failure.

## Desired behavior

- A successful continuation that reaches another conflict must yield a complete next conflict receipt and return to the bounded conflict-resolution loop.
- If either `git rebase --continue` or follow-up receipt capture fails, persist a bounded, redacted diagnostic that identifies the phase, command/return code, and receipt-capture failure.
- Do not collapse a known follow-up conflict into `host could not continue rebase`.
- Preserve the existing fail-closed ownership, edit-scope, remote-head, signing, and publication invariants.

## Acceptance criteria

- [ ] Add an integration-level regression using a real disposable Git repository whose rebase requires at least two sequential conflict resolutions.
- [ ] Prove the first host continuation returns a valid receipt for the second conflict and the stage schedules another resolver turn rather than terminalizing.
- [ ] Prove the second resolver/continuation completes and reaches the normal post-rebase validation/publish path.
- [ ] When follow-up receipt capture fails, return a structured failure kind plus bounded redacted stderr/stdout from the failed continuation and a separate bounded receipt-capture diagnostic.
- [ ] Persist the same safe diagnostic in the WorkItem/event/review surfaces without secrets or unbounded output.
- [ ] Cover remote-head drift, out-of-scope edits, unresolved markers, signing failure, and genuine `git rebase --continue` failure as distinct fail-closed outcomes.
- [ ] Existing focused rebase tests, Ruff, mypy, and the full test suite pass in a sanitized test environment.

## Evidence

- PR #2780 was manually rebased from `24b32c1` onto `main` at `7c62eb60`; Git completed after resolving the two sequential conflicts.
- The original loop failure retained only the generic host error, so the exact reason follow-up receipt capture failed cannot be reconstructed from durable state.
- The full local pre-push suite is independently affected by ambient guard-recovery configuration injected into config-merge tests; that pollution is out of scope for #2780 and should not be conflated with rebase recovery.

mvillmow:
Summary:
- reinject only validated SSH signing identity into controlled rebase continuation
- preserve bounded redacted Git diagnostics for signing and continuation failures
- distinguish follow-up conflicts from host signing failures

Validation:
- uv run pytest tests/unit -q
- uv run pytest --no-cov -q tests/unit/automation/pipeline/test_worker_pool.py tests/unit/utils/test_git.py
- uv run ruff check ...
- uv run mypy hephaestus/automation/pipeline/worker_pool.py hephaestus/utils/git.py

Closes #2783

### DIFF
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -78,13 +78,13 @@ from hephaestus.automation.worktree_manager import (
     BranchWorktreeOwnedError,
     WorktreeManager,
 )
+from hephaestus.diagnostics import bounded_git_diagnostic
 from hephaestus.io.utils import write_secure
 from hephaestus.resilience import (
     CircuitBreakerOpenError,
     resilient_call,
 )
 from hephaestus.utils.file_lock import LockUnavailableError, file_lock
-from hephaestus.utils.git import bounded_git_diagnostic
 from hephaestus.utils.helpers import get_repo_root
 
 logger = logging.getLogger(__name__)
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -1225,11 +1225,11 @@ def _parse_host_git_signing_config(raw: str) -> dict[str, str] | None:
 
 def _validated_signing_key(value: str) -> Path | None:
     """Resolve an absolute, private, regular SSH signing key path."""
-    signing_key = Path(value).expanduser()
     try:
+        signing_key = Path(value).expanduser()
         resolved_key = signing_key.resolve(strict=True)
         mode = resolved_key.stat().st_mode
-    except OSError:
+    except (OSError, RuntimeError, ValueError):
         return None
     if (
         not signing_key.is_absolute()
--- a/hephaestus/automation/pipeline/worker_pool.py
+++ b/hephaestus/automation/pipeline/worker_pool.py
@@ -84,6 +84,7 @@ from hephaestus.resilience import (
     resilient_call,
 )
 from hephaestus.utils.file_lock import LockUnavailableError, file_lock
+from hephaestus.utils.git import bounded_git_diagnostic
 from hephaestus.utils.helpers import get_repo_root
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


tests/unit/automation/pipeline/test_worker_pool.py
```diff
@@ -22,6 +22,7 @@
 
 import pytest
 
+import hephaestus.automation.pipeline.worker_pool as worker_pool_module
```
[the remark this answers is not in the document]
mvillmow: Fixed in f0341aee: the helper now uses the existing from-import block, removing the duplicate module import. Focused Ruff and regression tests pass, and the protected pre-push suite passed 7,906 tests.

hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -2666,37 +2769,73 @@ def _continue_rebase_process(
...
+            next_receipt: dict[str, object] | JobResult | None = None
+            if phase == "rebase_continue":
+                next_receipt = self._conflict_receipt(
+                    cwd,
+                    remote=remote,
+                    base_branch="main",
+                    expected_remote_sha=expected_remote_sha,
+                    timeout=timeout,
+                )
+                if isinstance(next_receipt, dict):
```
mvillmow, line 2802: **[major][required] Missing the required multi-conflict recovery regression.** This branch is the core path that must return a second conflict receipt, yet the PR adds only a mocked signing-failure test. Issue #2783 explicitly requires a real disposable-repository flow with two sequential conflicts, proving that the first continuation schedules a second resolver turn and the second continuation reaches normal completion/publish. Add that integration-level regression (and retain the receipt assertions), otherwise a broken transition between conflict turns can still pass this suite.

mvillmow (commented):
Architecture is aligned and all 33 head-bound checks pass, but the linked issue's recovery and durable-diagnostics acceptance criteria remain incomplete. Three required findings are attached.

hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -2666,37 +2769,73 @@ def _continue_rebase_process(
...
+                    next_receipt["base_sha"] = base_sha
+                    return JobResult(
+                        ok=False,
+                        value=next_receipt,
+                        error="rebase conflict resolution required: additional conflicts found",
+                    )
+            stdout_tail = bounded_git_diagnostic(exc.stdout, limit=_TAIL)
```
mvillmow, line 2809: **[major][required] The new diagnostics never reach the required durable surfaces.** Although this `JobResult` carries redacted tails and `receipt_error`, `ImplementationStage._rebase_continue_wait` only records a generic `rebase_error` and the coordinator's durable event records only that generic error. A terminal failure therefore remains `implementation_rebase_failed`, losing the continuation and receipt-capture diagnostics that #2783 requires in the WorkItem/event/review surfaces. Thread the bounded, already-redacted fields through the stage's persisted payload/event representation and add a regression that observes the terminal durable state.

hephaestus/automation/pipeline/worker_pool.py
```diff
@@ -1201,6 +1202,108 @@ def _controlled_git_env() -> dict[str, str]:
...
+        key, separator, value = entry.partition("\n")
+        if not separator or key not in _HOST_SIGNING_CONFIG_KEYS or key in parsed:
+            return None
+        parsed[key] = value
+    return parsed if set(parsed) == set(_HOST_SIGNING_CONFIG_KEYS) else None
+
+
+def _validated_signing_key(value: str) -> Path | None:
+    """Resolve an absolute, private, regular SSH signing key path."""
+    signing_key = Path(value).expanduser()
```
mvillmow: **[minor][required] Fail closed for malformed signing-key paths.** `Path.expanduser()` can raise `RuntimeError` for an unknown `~user`, and `resolve()` can raise `ValueError` for embedded NUL input, but this validator catches only `OSError`. A malformed global `user.signingkey` consequently escapes as a generic worker crash instead of returning the intended `signing_configuration` result. Catch those path-validation exceptions (or validate before constructing/resolving the path) and add malformed-config coverage.

mvillmow: Fixed in signed commit 626b227a. Signing-key validation now fails closed on OSError, RuntimeError, and ValueError, with regressions for an unknown tilde user and an embedded-NUL path. These malformed values now return the intended signing_configuration result instead of escaping as worker crashes.

hephaestus/automation/pipeline/worker_pool.py
mvillmow: Fixed in signed commit 1b479ff3. The new regression creates a disposable bare origin and checkout, produces two independent sequential conflicts, asserts the first continuation returns the second conflict receipt, then proves the second continuation publishes the signed/DCO head under the original exact lease. Full pre-push verification passed 7,911 tests at 85.23% coverage.

hephaestus/automation/pipeline/worker_pool.py
mvillmow: Fixed in signed commit 626b227a, with the zero-I/O boundary cleanup in signed commit dda57087. ImplementationStage now persists the bounded rebase diagnostic in the WorkItem payload, and CoordinatorRuntime emits the bounded/redacted diagnostic in durable job events. Regressions observe both surfaces; the affected 473-test suite and full 7,911-test pre-push gate pass.

hephaestus/utils/git.py
```diff
@@ -3,11 +3,14 @@
...
 import logging
-import re
 import subprocess
 from pathlib import Path
 from typing import Any
 
+from hephaestus.diagnostics import (
+    bounded_git_diagnostic as bounded_git_diagnostic,
+    redact_git_diagnostic as _redact_git_diagnostics,
+)
```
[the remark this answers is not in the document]
mvillmow: Fixed in signed commit 194e73cf. The compatibility re-export was removed, and the regression now imports bounded_git_diagnostic from its pure owning module directly. Ruff, mypy, the focused diagnostics/boundary suite, and the 7,911-test pre-push gate all pass.

PR #2792, opened 2026-08-19
fix(agents): discover selected Pi isolation adapters

### LINKED ISSUE #2791: Bootstrap explicitly selected Pi isolation adapters in fresh processes
Parent: #2519

Blocks: #2738

Related implementation: #2737

## Defect

The Pi isolation adapter can currently be registered only through the in-process test seam. A fresh automation process cannot discover an installed host adapter, so provider-executed Pi work remains fail closed even when a broker package is installed.

## Scope

Add only the neutral discovery/bootstrap contract. Broker implementation, deployment, OS-isolation claims, and Stage 6 evidence remain owned by #2738 and #2519.

## Acceptance criteria

- An operator explicitly selects one adapter factory by environment variable from the `hephaestus.pi_isolation_adapters` entry-point group.
- A fresh process loads exactly one matching zero-argument factory and validates the returned adapter before provider dispatch.
- Unselected, missing, duplicate, load-failing, initialization-failing, and invalid adapters fail closed with sanitized diagnostics.
- The existing in-process registration seam remains supported and takes precedence.
- Focused tests cover successful fresh-process bootstrap and every fail-closed boundary without requiring a real broker.
- Runtime and architecture documentation describe selection, ownership, and the boundary that discovery is not broker or end-to-end evidence.

mvillmow:
## Summary

- load only the explicitly selected `hephaestus.pi_isolation_adapters` factory in fresh processes
- preserve stock fail-closed behavior and in-process registration precedence
- bootstrap the selected adapter before direct policy dispatch
- sanitize discovery, initialization, and protocol-attribute failures without leaking private diagnostics

## Root cause

The existing adapter registration seam worked only inside an embedding Python process. Installed console commands start fresh processes, leaving the external broker in #2738 impossible to select without mixing its discovery contract into the blocked Stage 6 PR.

## Prerequisite

The platform-aware host-verification receipt contract discovered during review was isolated into #2793 and merged through PR #2794. This branch is rebased onto that merge; unsupported platforms now produce exact-head `status: skipped` N/A receipts without executing PR code, while supported-host failures remain fail closed.

## Validation

- Pi runtime, execution-policy, plugin, session, confinement, ADR, and documentation suites: 197 passed, 1 skipped
- Ruff check and format validation passed
- mypy passed for the changed Python files
- configured pre-commit checks passed for every changed path

Closes #2791

### DIFF
--- a/hephaestus/agents/runtime.py
+++ b/hephaestus/agents/runtime.py
@@ -263,11 +263,12 @@ def _load_configured_pi_isolation_adapter() -> None:
     try:
         factory = matches[0].load()
         adapter = factory()
+        invoke = getattr(adapter, "invoke", None)
     except Exception:
         raise PiIsolationUnavailableError(
             f"Pi isolation adapter {adapter_name!r} could not be initialized"
         ) from None
-    if not callable(getattr(adapter, "invoke", None)):
+    if not callable(invoke):
         raise PiIsolationUnavailableError(
             f"Pi isolation adapter {adapter_name!r} does not implement invoke()"
         )
--- a/hephaestus/agents/runtime.py
+++ b/hephaestus/agents/runtime.py
@@ -1938,6 +1938,8 @@ def _run_pi_with_policy(
     """
     command = _pi_base_cmd(session_id=session_id)
     command.extend(_pi_policy_args(policy))
+    if _PI_ISOLATION_ADAPTER is None:
+        _load_configured_pi_isolation_adapter()
     adapter = _PI_ISOLATION_ADAPTER
     if adapter is None:
         raise PiIsolationUnavailableError(
--- a/hephaestus/agents/runtime.py
+++ b/hephaestus/agents/runtime.py
@@ -17,6 +17,7 @@ import time
 from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass
 from enum import StrEnum
+from importlib.metadata import entry_points
 from pathlib import Path
 from typing import Any, Literal, Protocol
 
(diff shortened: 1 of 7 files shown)


hephaestus/agents/runtime.py
```diff
@@ -236,8 +239,45 @@ def register_pi_isolation_adapter(adapter: PiIsolationAdapter) -> None:
...
+        raise PiIsolationUnavailableError(
+            f"Pi isolation adapter {adapter_name!r} does not implement invoke()"
+        )
+    register_pi_isolation_adapter(adapter)
+
+
 def _require_pi_isolation_adapter() -> None:
     """Fail at provider selection when this installation has no Pi broker."""
+    if _PI_ISOLATION_ADAPTER is None:
+        _load_configured_pi_isolation_adapter()
```
mvillmow, line 281: [Review] The selected entry point is loaded only through `resolve_agent("pi")`, but the direct provider paths (`run_agent_text`, `run_agent_session`, and resume) dispatch through `_run_pi_with_policy`, which still reads `_PI_ISOLATION_ADAPTER` directly. A fresh process with `HEPH_PI_ISOLATION_ADAPTER` set can therefore fail with the old unavailable-adapter error if it calls the direct runner without a prior resolve, so the new bootstrap is not guaranteed before provider dispatch. Move the load/validation check into `_run_pi_with_policy` or call `_require_pi_isolation_adapter()` immediately before reading the global adapter, and add env-only dispatch coverage.

tests/unit/agents/test_execution_policy.py
```diff
@@ -176,12 +181,233 @@ def invoke(self, **_kwargs: object) -> agent_runtime.AgentRunResult:
...
+
+    monkeypatch.setattr(
+        agent_runtime, "preflight_pi_environment", lambda _cwd: PiPreflightResult.ready_result()
+    )
+    monkeypatch.setattr(agent_runtime, "is_agent_authenticated", lambda _agent: True)
+    monkeypatch.setattr(agent_runtime, "_PI_ISOLATION_ADAPTER", None)
+    monkeypatch.setattr(agent_runtime, "entry_points", entry_points, raising=False)
+    monkeypatch.setenv("HEPH_PI_ISOLATION_ADAPTER", "operator-broker")
+
+    assert agent_runtime.resolve_agent("pi", cwd=tmp_path) == "pi"
```
mvillmow, line 223: [Review] The review-supplied validation receipts for this added bootstrap path are all `ok:false` with `unsupported_host_verification_boundary`, including the focused pytest receipt for this file, so the PR currently has no host-bound execution evidence that these tests pass on the reviewed head. Please provide passing reviewed-head receipts for the fixed validation plan before treating this as validated.

mvillmow: [Response] Verified the reviewed-head Pi bootstrap path with a passing focused run (`tests/unit/agents/test_execution_policy.py`: 22 passed) in the local workspace venv; no code or doc change was needed for this evidence-only thread.

[auto-msg] reply has no corresponding commit, review thoroughly

PR #2804, opened 2026-08-21
fix(ci): run required checks at publication and merge queue gates

### LINKED ISSUE #2802: fix(ci): run actual required checks before publication and in merge queue
## Problem

Hephaestus has two gaps in its required-check execution:

1. The automation implementation queue does not reliably run the repository-owned source-check suite at the initial publication boundary. A generic opt-in pytest command can drift from `.github/workflows/_required.yml`.
2. GitHub merge-queue commits run only `merge-queue-smoke.yml`. The workflows that produce the repository's actual required contexts (`_required.yml` and `test.yml`) do not accept `merge_group`, so the synthetic queue commit is not validated by the required suite.

## Desired behavior

The implementation queue runs the canonical local source-check suite once immediately before the initial commit/push/PR publication boundary. A failure returns the item to the implementer and the suite must pass after the fix before publication. Existing-PR remediation does not repeatedly run the automatic full local suite.

GitHub Actions runs the real required-context workflows for every `merge_group: checks_requested` event. The separate queue-smoke substitute is retired.

## Acceptance criteria

- A fresh Hephaestus implementation automatically runs the fixed trusted command `env HEPHAESTUS_CI_REBUILD=1 bash scripts/run_ci_local.sh all` before commit, push, and PR creation.
- A failed local run returns to the implementer test-fix path and is rerun before initial publication.
- A passing local run advances to commit/push and PR creation without repeating the suite.
- Existing-PR remediation skips the automatic full local gate; other repositories preserve their configurable opt-in behavior.
- `scripts/run_ci_local.sh all` remains aligned with locally executable source checks in `_required.yml`; PR-metadata-only policy remains GitHub-side.
- `_required.yml` and `test.yml` both run on `merge_group: checks_requested`, so queue commits emit the aggregate, direct-ruleset, and classic matrix contexts.
- The smoke-only merge-queue workflow is removed.
- Regression tests cover the local red/fix/retest/publication route and required workflow triggers.

mvillmow:
## Summary

- run the repository-owned pre-PR gate once before initial publication and route failures back to implementation
- validate the exact candidate bytes, including non-ignored untracked files, without mutating the real Git index or object database
- run the actual required workflows on merge-group commits and revalidate every source PR policy result under HEADGREEN
- make merge-group license violations blocking and align required shell jobs with the version-pinned local CI image
- lock workflow, context, aggregate behavior, and operator documentation with regression contracts

## Verification

- exact pushed head: `f0e9b0bd5a8017e767e3459cc9cd105cd7673aad`
- pre-push repository gate: 7,967 passed, 13 skipped, 7 deselected; 85.26% coverage
- all pre-commit hooks passed, including Ruff, MyPy, ShellCheck, workflow schema, Bandit, zizmor, and lock validation
- focused queue-policy and local-runner CI tests: 35 passed; complete `tests/unit/ci`: 234 passed
- the offline workflow-schema invocation passed with `UV_NO_SYNC=1 UV_OFFLINE=1`
- exact-head Required Checks run `32537796452` passed every job and the aggregate `required-checks-gate`; CodeQL and both classic Test contexts also passed for the same SHA
- the pushed commit has a valid ED25519 Git signature and a DCO `Signed-off-by` trailer
- final review re-collects live GitHub evidence rather than treating this prose as authority

Closes #2802


hephaestus/automation/pipeline/stages/implementation.py
```diff
@@ -994,19 +1009,34 @@ def _test_wait(self, item: WorkItem, ctx: StageContext) -> StepResult:
             # on_job_done (doc: agent_error consumes the implement
             # budget); RETRY re-enters the stage for the next attempt.
             return StageOutcome(Disposition.RETRY, "agent_error")
-        if not getattr(ctx.config, "run_pre_pr_tests", False):
+        is_hephaestus = item.repo.rsplit("/", 1)[-1].casefold() == "hephaestus"
```
mvillmow: **major · required** — The automatic profile is selected by repository basename only. Because `ctx.org` is available, an unrelated `OtherOrg/Hephaestus` is forced to run HomericIntelligence’s hardcoded script and loses its configurable opt-in behavior, violating #2802’s acceptance criterion for other repositories. Match the full canonical organization/repository identity and add a cross-organization same-name regression.

hephaestus/automation/pipeline/stages/implementation.py
```diff
@@ -994,19 +1009,34 @@ def _test_wait(self, item: WorkItem, ctx: StageContext) -> StepResult:
...
+        run_hephaestus_pre_pr_checks = (
```
mvillmow, line 1016: **major · required** — This gate runs before `commit_push`, against a dirty worktree, but the suite does not cover the exact candidate bytes. `pre-commit --all-files` resolves `git ls-files` and excludes new untracked files; Gitleaks `detect --source=.` scans Git history, while the later commit path stages ordinary new files. New lint violations or secrets can therefore pass the local gate and be published before Actions catches them. Bind/stage the candidate manifest and scan staged/directory bytes, or create a local candidate commit and verify that exact SHA before push; failures should still return to `TESTFIX_WAIT`.

.github/workflows/_required.yml
```diff
@@ -16,10 +16,8 @@ on:
       - ready_for_review
   push:
     branches: [main]
-  # merge_group intentionally removed: merge-queue checks are served by the
-  # single fast `merge-queue-smoke` job in merge-queue-smoke.yml. Re-running
-  # all 16+ jobs per queue entry serialized on runner slots (70-90 min per
-  # merge). PR-side jobs in this workflow are unchanged.
+  merge_group:
```
mvillmow, line 19: **major · required** — Enabling `merge_group` activates the existing `pr-policy` merge-group path, but that path only emits an unconditional `echo`; checkout, metadata retrieval, and all three policy validations remain `pull_request`-only. The live ruleset uses `HEADGREEN`, under which only the group head must pass and failing entries may merge with a passing entry ([GitHub rules API](https://docs.github.com/en/rest/repos/rules?apiVersion=2026-03-10)). This required context is therefore not the actual policy check. Revalidate every constituent PR before reporting success, or bind the queue to an `ALLGREEN` policy and prove that invariant.

mvillmow, line 19: **major · required** — This trigger also weakens `license-scan`: the workflow forwards `GITHUB_EVENT_NAME=merge_group`, while `scripts/check_license_compatibility.py` returns failure for violations only when the value is `pull_request`; every other event is advisory exit 0. The aggregate can therefore report green for a violating synthetic queue commit. Treat `merge_group` as blocking (prefer an explicit blocking mode) and add a violation regression for that event.

tests/unit/ci/test_merge_queue_required_checks.py
```diff
@@ -0,0 +1,28 @@
...
+
+def _load_workflow(path: Path) -> dict[str, Any]:
+    """Load workflow YAML without YAML 1.1 coercing the ``on`` key to true."""
+    return cast(
+        dict[str, Any],
+        yaml.load(path.read_text(encoding="utf-8"), Loader=yaml.BaseLoader),
+    )
+
+
+def test_every_required_context_workflow_runs_for_merge_groups() -> None:
```
mvillmow, line 43: **minor · required** — Despite its name, this test proves only that both workflow files contain a `merge_group` trigger. It cannot detect renamed/missing required contexts, changed matrix names, incomplete aggregate membership, or event branches that report vacuous success. Assert the exact documented classic/ruleset context mapping, complete aggregate `needs`, and blocking `merge_group` behavior for policy/license gates.

ci/Containerfile
```diff
@@ -149,11 +149,13 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # Runtime-only system packages (no build tools), resolved from the same
 # immutable snapshot as the builder stage.
 RUN apt-get update && apt-get install -y --no-install-recommends \
+    bats \
```
mvillmow: **minor · required** — The local gate now takes Bats and ShellCheck from the pinned Debian snapshot, while required CI installs them from Ubuntu 24.04. Those cross-distribution package versions are not equivalent, so the local result can disagree with the required job even with identical flags. Run the required jobs through this same image or pin the same tool artifacts on both paths, and make the contract tests verify versions rather than command strings alone.

.github/workflows/README.md
```diff
@@ -9,7 +9,6 @@
 | auto-label-severity.yml | Reconciles `severity:*` labels from issue form content. | issues opened/edited |
 | auto-tag.yml | Manually computes and pushes the next signed release tag. | workflow_dispatch |
 | contract.yml | Opt-in authenticated GitHub and agent contract tests. | workflow_dispatch |
-| merge-queue-smoke.yml | Single fast `merge-queue-smoke` job for merge-queue entries (lockfile, workflow-schema, and symlink validation). | merge_group |
```
mvillmow, line 12: **minor · required** — Removing the smoke-workflow row leaves the inventory claiming that `_required.yml` and `test.yml` trigger only on pull requests and pushes, even though this change adds `merge_group` to both workflows. Update both trigger cells so the operator-facing workflow contract matches what now executes.

mvillmow (commented):
Architecture is directionally aligned with ADR-0007, and the 12 exact live required contexts match the workflow job names. The behavioral gate is not yet equivalent on the pre-publication and merge-queue paths.

Score: Architecture 25/30; Issue/scope 15/20; Implementation 12/18; Testing/evidence 10/15; Security/safety 8/10; Integration/release 6/7 — 76/100 (C).

Verdict: NO-GO. Four major and three minor required findings are attached.

Binding: PR #2804, base e9976c1ef27ba823b04f5b5c6c48ee95cb139f8a, head 6048761e3ac51e5470c8185f69d662551e3e6388, 18-file manifest 27bd817940af195b58295e03dfc16fb32b95b8ee3a57aa9d6bb4b8aaf3c7c626, issue #2802, behind 0. All 33 head-bound PR check runs are complete and all 12 currently required contexts are green; that evidence does not exercise a merge-group event. Auto-merge: not eligible.

hephaestus/automation/pipeline/stages/implementation.py
mvillmow: Fixed in 6b1a90ce. The local runner now constructs an alternate Git index from HEAD plus all non-ignored working-tree changes, including untracked files, without changing the real index. Pre-commit uses that candidate index, and Gitleaks scans both history and the materialized candidate tree. Regressions cover untracked lint and secret inputs plus real-index preservation.

.github/workflows/_required.yml
mvillmow: Fixed in 6b1a90ce using the source-PR evidence path. On a merge-group event, the pr-policy job parses the queue ref, fetches that open PR current head, and fails unless every latest pr-policy run from GitHub Actions app 15368 is completed successfully on that exact SHA. Missing, malformed, stale, pending, or failed evidence is blocking. The query was also exercised read-only against a live queued PR.

ci/Containerfile
mvillmow: Fixed in 6b1a90ce. Required ShellCheck and Bats jobs now build and run the same hephaestus-ci:local image as the local gate. The Containerfile pins ShellCheck 0.10.0-1 and Bats 1.11.1-1 in the runtime stage and verifies their reported versions while building.

hephaestus/automation/pipeline/stages/implementation.py
mvillmow: Fixed in 6b1a90ce. ImplementationStage._test_wait now matches the full configured organization/repository identity. The regression verifies that OtherOrg/Hephaestus keeps its configurable pre-PR command.

.github/workflows/README.md
mvillmow: Fixed in 6b1a90ce. The workflow inventory now lists merge-group events for both _required.yml and test.yml.

tests/unit/ci/test_merge_queue_required_checks.py
mvillmow: Fixed in 6b1a90ce. The workflow contract tests now assert the documented classic matrix contexts, direct ruleset context names, complete aggregate dependencies, source-PR policy revalidation, blocking merge-group licensing, and shared-image shell jobs.

.github/workflows/_required.yml
mvillmow: Fixed in 6b1a90ce. License violations now return failure for both pull-request and merge-group events; the regression supplies a violating merge-group event and requires exit status 1.

